PR #1, opened 2021-11-01
Authorino Reconciler

jjaferson:
This PR is the starting point to enable the deployment of [Authorino](https://github.com/Kuadrant/authorino) instances via the operator.

Ps.: missing unit testing

## Verification steps

1) Clone the repo from this branch
2) Install the operator on a cluster of your choice
3) Run the operator eg: `make install run`
4) Create an instance of authorino using the CR, an example can be [found here](https://github.com/Kuadrant/authorino-operator/blob/foundation2/config/samples/authorino-operator_v1beta1_authorino.yaml).
5) Check if the authorino instance was deployed correctly by verifying whether a deployment is created and a pod is up and running


config/samples/authorino-operator_v1beta1_authorino.yaml
```diff
@@ -0,0 +1,19 @@
...
+  image: quay.io/3scale/authorino:latest
+  replicas: 1
+  imagePullPolicy: Always
+  clusterWide: true
+  listener:
+    port: 
+    tsl: true
+    certPath: ""
+    certKeyPath: ""
+  oidcserver:
```
guicassolato, line 16:
```suggestion
  oidcServer:
```

api/v1beta1/authorino_types.go
```diff
@@ -0,0 +1,139 @@
...
+	AuthorinoOperatorNamespace string        = "authorino-operator"
+
+	// Authorino EnvVars
+	ExtAuthGRPCPort         string = "EXT_AUTH_GRPC_PORT"
+	TSLCertPath             string = "TLS_CERT"
+	TSLCertKeyPath          string = "TLS_CERT_KEY"
+	OIDCHTTPPort            string = "OIDC_HTTP_PORT"
+	OIDCTSLCertPath         string = "OIDC_TLS_CERT"
+	OIDCTLSCertKeyPath      string = "OIDC_TLS_CERT_KEY"
+	AuthConfigLabelSelector string = "AUTH_CONFIG_LABEL_SELECTOR"
```
guicassolato, line 42: Attention to https://github.com/Kuadrant/authorino/pull/179, changing the name of this env var to `SECRET_LABEL_SELECTOR`, and introducing `AUTH_CONFIG_LABEL_SELECTOR`.

api/v1beta1/authorino_types.go
```diff
@@ -0,0 +1,139 @@
...
+	SecretLabelSelector     string = "SECRET_LABEL_SELECTOR"
+)
```
guicassolato, line 44: We should include `LOG_LEVEL` (`error`, `info`, `debug`; `default: `info`) and `LOG_MODE` (`production`, `development`; `default: `production`)

main.go
```diff
@@ -67,13 +71,21 @@ func main() {
 		Port:                   9443,
 		HealthProbeBindAddress: probeAddr,
 		LeaderElection:         enableLeaderElection,
-		LeaderElectionID:       "f0fb677d.authorino.3scale.net",
+		LeaderElectionID:       "aac3a15d.kuadrant.3scale.net",
```
guicassolato, line 74:
```suggestion
		LeaderElectionID:       "aac3a15d.authorino.kuadrant.net",
```

api/v1beta1/groupversion_info.go
```diff
@@ -0,0 +1,36 @@
...
+package v1beta1
+
+import (
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/scheme"
+)
+
+var (
+	// GroupVersion is group version used to register these objects
+	GroupVersion = schema.GroupVersion{Group: "authorino-operator.kuadrant.3scale.net", Version: "v1beta1"}
```
guicassolato, line 29:
```suggestion
	GroupVersion = schema.GroupVersion{Group: "authorino-operator.authorino.kuadrant.net", Version: "v1beta1"}
```

controllers/authorino_controller.go
```diff
@@ -0,0 +1,598 @@
...
+// the user.
+//
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.9.2/pkg/reconcile
+func (r *AuthorinoReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := r.Log.WithValues("authorinoReconciler", req.NamespacedName)
+
+	//get operator namespace
+	r.OperatorNamespace = authorinooperatorv1beta1.AuthorinoOperatorNamespace
+	ns, found := os.LookupEnv("WATCH_NAMESPACE")
```
guicassolato, line 69: I'd rather have all the envs loaded in `main.go` and passes to `AuthorinoReconciler` in a field.

config/crd/bases/authorino-operator.kuadrant.3scale.net_authorinoes.yaml
```diff
@@ -0,0 +1,133 @@
...
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.6.1
+  creationTimestamp: null
+  name: authorinoes.authorino-operator.kuadrant.3scale.net
+spec:
+  group: authorino-operator.kuadrant.3scale.net
+  names:
+    kind: Authorino
+    listKind: AuthorinoList
+    plural: authorinoes
```
guicassolato, line 15: I wonder whether a more natural spelling wouldn't be "authorinos".

(BTW I wish I could pluralize it as "authorini", but people would just kill me 🙂)

api/v1beta1/groupversion_info.go
guicassolato: Ref: https://github.com/Kuadrant/authorino/issues/173

config/authorino/rbac/role.yaml
```diff
@@ -0,0 +1,45 @@
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  creationTimestamp: null
+  name: manager-role
```
guicassolato, line 7: I think this role name will collide with the operator's `manager-role`.

config/authorino/rbac/service_editor_role.yaml
```diff
@@ -0,0 +1,23 @@
+# permissions for end users to edit services.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: service-editor-role
```
guicassolato, line 5: Perhaps...

```suggestion
  name: authconfig-editor-role
```

WDYT?

config/authorino/rbac/service_viewer_role.yaml
```diff
@@ -0,0 +1,20 @@
+# permissions for end users to view services.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: service-viewer-role
```
guicassolato, line 5: Same as https://github.com/Kuadrant/authorino-operator/pull/1/files#r740874424

```suggestion
  name: authconfig-viewer-role
```

main.go
```diff
@@ -67,13 +71,21 @@ func main() {
...
 	})
 	if err != nil {
 		setupLog.Error(err, "unable to start manager")
 		os.Exit(1)
 	}
 
+	if err = (&controllers.AuthorinoReconciler{
+		Client: mgr.GetClient(),
+		Log:    ctrl.Log.WithName("authorino-operator").WithName("controller").WithName("Authorino"),
```
guicassolato, line 83: We could use github.com/kuadrant/authorino/pkg/common/log.

controllers/authorino_controller.go
```diff
@@ -0,0 +1,598 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	"github.com/go-logr/logr"
```
guicassolato, line 37:
```suggestion
	"github.com/go-logr/logr"
	
```

controllers/authorino_controller.go
```diff
@@ -0,0 +1,598 @@
...
+// move the current state of the cluster closer to the desired state.
+// TODO(user): Modify the Reconcile function to compare the state specified by
+// the Authorino object against the actual cluster state, and then
+// perform operations to make the cluster state reflect the state specified by
+// the user.
+//
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.9.2/pkg/reconcile
+func (r *AuthorinoReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := r.Log.WithValues("authorinoReconciler", req.NamespacedName)
```
guicassolato, line 65:
```suggestion
	log := r.Log.WithValues("authorino", req.NamespacedName)
```

As `req.NamespacedName` is the name of the resource (not the name of the reconciler) and the resource kind is "Authorino".

Refer to https://github.com/kubernetes-sigs/controller-runtime/blob/master/TMP-LOGGING.md for more info.

controllers/authorino_controller.go
```diff
@@ -0,0 +1,598 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	"github.com/go-logr/logr"
+	authorinooperatorv1beta1 "github.com/kuadrant/authorino-operator/api/v1beta1"
+)
+
+// AuthorinoReconciler reconciles a Authorino object
+type AuthorinoReconciler struct {
+	client.Client
+	OperatorNamespace string
```
guicassolato, line 44:
```suggestion
	Namespace string
```

controllers/authorino_controller.go
```diff
@@ -0,0 +1,598 @@
...
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *AuthorinoReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&authorinooperatorv1beta1.Authorino{}).
+		Complete(r)
+}
+
+func (r *AuthorinoReconciler) authorinoInstance(namespacedName types.NamespacedName) (*authorinooperatorv1beta1.Authorino, error) {
```
guicassolato, line 157:
```suggestion
func (r *AuthorinoReconciler) getAuthorinoInstance(namespacedName types.NamespacedName) (*authorinooperatorv1beta1.Authorino, error) {
```

controllers/authorino_controller.go
```diff
@@ -0,0 +1,598 @@
...
+	}
+
+	// get authorino instance
+	authorinoInstance, err := r.authorinoInstance(req.NamespacedName)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if authorinoInstance == nil {
+		log.Info("Authorino CR not found. returning the reconciler")
```
guicassolato, line 81: This is a bit tricky. 404 can return as an error.

Check how we're handling in https://github.com/Kuadrant/authorino/blob/d5e243f5172d95a1f5b850e92a0d29e7df78780c/controllers/auth_config_controller.go#L63-L71 for an example.

controllers/authorino_controller.go
```diff
@@ -0,0 +1,598 @@
...
+		return ctrl.Result{}, nil
+	}
+
+	log.Info("Found an instance of authorino", "authorinoInstanceName", authorinoInstance.Name)
```
guicassolato, line 85:
```suggestion
	log.V(1).Info("Found an instance of authorino", "authorinoInstanceName", authorinoInstance.Name)
```

controllers/authorino_controller.go
```diff
@@ -0,0 +1,598 @@
...
+		},
+		{
+			APIGroups: []string{"coordination.k8s.io"},
+			Resources: []string{"leases"},
+			Verbs:     []string{"get", "list", "create", "update"},
+		},
+	}
+}
+
+func (r *AuthorinoReconciler) authorinoServices(authorino *authorinooperatorv1beta1.Authorino) error {
```
guicassolato, line 534:
```suggestion
func (r *AuthorinoReconciler) createOrUpdateAuthorinoServices(authorino *authorinooperatorv1beta1.Authorino) error {
```

Let's favour function names that let more explicit what the function does.

controllers/authorino_controller.go
```diff
@@ -0,0 +1,598 @@
...
+		log.Info("Authorino CR not found. returning the reconciler")
+		return ctrl.Result{}, nil
+	}
+
+	log.Info("Found an instance of authorino", "authorinoInstanceName", authorinoInstance.Name)
+	authorinoInstance.Status.Ready = false
+
+	err = r.authorinoServices(authorinoInstance)
+	if err != nil {
+		authorinoInstance.Status.Ready = false
```
guicassolato, line 90: You'll probably need to call `r.Status().Update(ctx, authorinoInstance)` to persist this.

controllers/authorino_controller.go
```diff
@@ -0,0 +1,598 @@
...
+		return ctrl.Result{}, err
+	}
+
+	existingDeployment := &appsv1.Deployment{}
+	err = r.Get(context.TODO(),
+		types.NamespacedName{
+			Name:      authorinoInstance.GetName(),
+			Namespace: authorinoInstance.GetNamespace(),
+		}, existingDeployment)
+	if err != nil && errors.IsNotFound(err) {
```
guicassolato, line 108: I think this could be moved to function as well

```go
func (r *AuthorinoReconciler) getAuthorinoDeployment(authorino *authorinooperatorv1beta1.Authorino) (*appsv1.Deployment, error)
  deployment := &appsv1.Deployment{}
  err = r.Get(context.TODO(), types.NamespacedName{
    Name:      authorino.GetName(),
    Namespace: authorino.GetNamespace(),
  }, deployment)
  return deployment, err
```

And then the `if` statement

```go
if existingDeployment, err := getAuthorinoDeployment(authorinoInstance); err != nil && errors.IsNotFound(err)  {
  // create the deployment
} else if existingDeployment != nil {
  // update the deployment
} else {
  // error
}
```

controllers/authorino_controller.go
```diff
@@ -0,0 +1,598 @@
...
+		if errors.IsNotFound(err) {
+			r.Log.Info("Authorino CR not found.")
+			return nil, nil
+		}
+		return nil, err
+	}
+	return authorinoInstance, nil
+}
+
+func (r *AuthorinoReconciler) authorinoDeployment(authorino *authorinooperatorv1beta1.Authorino) *appsv1.Deployment {
```
guicassolato, line 170: Maybe...

```suggestion
func (r *AuthorinoReconciler) buildDesiredAuthorinoDeployment(authorino *authorinooperatorv1beta1.Authorino) *appsv1.Deployment {
```

?

controllers/authorino_controller.go
```diff
@@ -0,0 +1,598 @@
...
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	"github.com/go-logr/logr"
+	authorinooperatorv1beta1 "github.com/kuadrant/authorino-operator/api/v1beta1"
```
guicassolato, line 38: I'd be fine with either

```suggestion
	operator "github.com/kuadrant/authorino-operator/api/v1beta1"
```

or 

```suggestion
	api "github.com/kuadrant/authorino-operator/api/v1beta1"
```

If it doesn't specify further, then it's ours; otherwise we make it more explicit, such as `k8sapps "k8s.io/api/apps/v1"`, `k8score "k8s.io/api/core/v1"`, etc.

WDYT?

controllers/authorino_controller.go
```diff
@@ -0,0 +1,598 @@
...
+			authorinoInstance.Status.Ready = false
+			log.Error(err, "Failed to create Authorino deployment resource", newDeployment.Name, newDeployment.Namespace)
+			return ctrl.Result{}, err
+		}
+		// Deployment created successfully - return and requeue
+		return ctrl.Result{Requeue: true}, nil
+	} else if err != nil {
+		authorinoInstance.Status.Ready = false
+		log.Error(err, "Failed to get Deployment for Authorino")
+		return ctrl.Result{}, err
```
guicassolato, line 121: Same as https://github.com/Kuadrant/authorino-operator/pull/1/files#r740896089. I think you won't get to `handleStatusUpdate(authorinoInstance)` if you return here.

controllers/authorino_controller.go
```diff
@@ -0,0 +1,598 @@
...
+	}
+
+	authorino.Status.AuthorinoInstances = len(authorinoInstances.Items)
+	return nil
+}
+
+func labelsForAuthorino(name string) map[string]string {
+	return map[string]string{
+		"control-plane":     "controller-manager",
+		"authorino_cr_name": name,
```
guicassolato, line 596:
```suggestion
		"authorino-resource": name,
```

controllers/authorino_controller.go
```diff
@@ -0,0 +1,598 @@
...
+}
+
+func (r *AuthorinoReconciler) handleStatusUpdate(authorino *authorinooperatorv1beta1.Authorino) error {
+	authorinoInstances := authorinooperatorv1beta1.AuthorinoList{}
+	err := r.List(context.TODO(), &authorinoInstances)
+	if err != nil {
+		return fmt.Errorf("Failed getting authorino instances, err: %w", err)
+	}
+
+	authorino.Status.AuthorinoInstances = len(authorinoInstances.Items)
```
guicassolato, line 589: I don't get this field. Why would an Authorino instance hold the count of how many Authorino instances there are running in the space?

api/v1beta1/authorino_types.go
jjaferson: I left they both there to make it backwards compatible with previous version authorino, do you reckon we should remove this?

api/v1beta1/authorino_types.go
jjaferson: I'll create an issue to add the log level and log mode to the authorino CR API

controllers/authorino_controller.go
jjaferson: Makes sense

api/v1beta1/authorino_types.go
guicassolato: I think it depends on how we're going to tie versions of the operator to versions of the service, but I guess leaving both env vars is good enough for now.

Let's add a comment
```go
// TODO: remove after Authorino v0.5.0
```

config/crd/bases/authorino-operator.kuadrant.3scale.net_authorinoes.yaml
jjaferson: I think it's the default way that the `operator-sdk` pluralize the name of the objects, I am happy to change to  `authorinos` if it's not going against any standards.

config/authorino/rbac/role.yaml
jjaferson: kustomizer will add an `authorino` prefix name to the resource. https://github.com/Kuadrant/authorino-operator/blob/foundation2/config/authorino/kustomization.yaml#L12

config/authorino/rbac/service_editor_role.yaml
jjaferson: I changed the name of the yaml file too

main.go
jjaferson: Yup, I'll create an issue to add the same log package that Authorino uses

controllers/authorino_controller.go
jjaferson: The 404 error is already handled in the `getAuthorinoInstance` function, it checks if the resource is not found and if so it returns `nil` https://github.com/Kuadrant/authorino-operator/blob/foundation2/controllers/authorino_controller.go#L161

controllers/authorino_controller.go
jjaferson: I agree

controllers/authorino_controller.go
jjaferson: I'll create an issue to better handle the status update based on the conditions conventions https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties

controllers/authorino_controller.go
jjaferson: yeah, it doesn't make any sense I was trying to fill up the status block with more things, but as I mentioned above I will create an issue to improve the handling of the status block using conditions. 

I will remove this field.

Makefile
```diff
@@ -107,7 +106,7 @@ docker-push: ## Push docker image with the manager.
 
 ##@ Deployment
 
-install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~/.kube/config.
+install: manifests kustomize install-authorino ## Install CRDs into the K8s cluster specified in ~/.kube/config.
 	$(KUSTOMIZE) build config/crd | kubectl apply -f -
 
 uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified in ~/.kube/config.
```
guicassolato, line 112: Should we also uninstall Authorino?

api/v1beta1/groupversion_info.go
```diff
@@ -0,0 +1,36 @@
...
+package v1beta1
+
+import (
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/scheme"
+)
+
+var (
+	// GroupVersion is group version used to register these objects
+	GroupVersion = schema.GroupVersion{Group: "operator.authorino.kuadrant.net", Version: "v1beta1"}
```
guicassolato, line 29: Sorry, @jjaferson. It should be [`*.kuadrant.io`](https://github.com/Kuadrant/kuadrant-controller/blob/1d691f6a792cfc4bc0f39de7de3e6c2beba5488c/apis/networking/v1beta1/groupversion_info.go#L29).

```suggestion
	GroupVersion = schema.GroupVersion{Group: "operator.authorino.kuadrant.io", Version: "v1beta1"}
```

guicassolato (changes requested):

main.go
```diff
@@ -67,13 +71,29 @@ func main() {
...
+	ns, found := os.LookupEnv("WATCH_NAMESPACE")
+	if found {
+		namespace = ns
+	}
+
+	if err = (&controllers.AuthorinoReconciler{
+		Client:    mgr.GetClient(),
+		Log:       ctrl.Log.WithName("authorino-operator").WithName("controller").WithName("Authorino"),
+		Scheme:    mgr.GetScheme(),
+		Namespace: namespace,
```
guicassolato, line 92: My understanding for "`WATCH_NAMESPACE`" is that this should be the namespace the operator will be watching for `Authorino` CRs (if not the entire cluster). Instead, it here seems to be used as the namespace where the operator is running maybe?

We here seems to be looking for a `ClusterRole` in the namespace that matches the value of this variable, with default to "`authorino-operator`", which doesn't sound right to me. We probably don't even need the name of the namespace to get the `ClusterRole`.

Maybe rename this variable... and either use `WATCH_NAMESPACE` as expected (an option of the manager, not of the controller) or make the operator always cluster-scoped (i.e. get rid of `WATCH_NAMESPACE`).

guicassolato: Here's what I had in mind:

```diff
diff --git a/api/v1beta1/authorino_types.go b/api/v1beta1/authorino_types.go
index 93febeb..38abcff 100644
--- a/api/v1beta1/authorino_types.go
+++ b/api/v1beta1/authorino_types.go
@@ -28,9 +28,8 @@ type ConditionType string
 
 const (
 	// ConditionReady specifies that the resource is ready
-	ConditionReady             ConditionType = "Ready"
-	AuthorinoContainerName     string        = "authorino"
-	AuthorinoOperatorNamespace string        = "authorino-operator"
+	ConditionReady         ConditionType = "Ready"
+	AuthorinoContainerName string        = "authorino"
 
 	// Authorino EnvVars
 	ExtAuthGRPCPort         string = "EXT_AUTH_GRPC_PORT"
diff --git a/controllers/authorino_controller.go b/controllers/authorino_controller.go
index 4163185..52d2d68 100644
--- a/controllers/authorino_controller.go
+++ b/controllers/authorino_controller.go
@@ -40,9 +40,8 @@ import (
 // AuthorinoReconciler reconciles a Authorino object
 type AuthorinoReconciler struct {
 	client.Client
-	Namespace string
-	Log       logr.Logger
-	Scheme    *runtime.Scheme
+	Log    logr.Logger
+	Scheme *runtime.Scheme
 }
 
 //+kubebuilder:rbac:groups=operator.authorino.kuadrant.net,resources=authorinoes,verbs=get;list;watch;create;update;patch;delete
@@ -329,8 +328,7 @@ func (r *AuthorinoReconciler) createAuthorinoPermission(authorino *api.Authorino
 
 	authorinoClusterRole := &k8srbac.ClusterRole{
 		ObjectMeta: v1.ObjectMeta{
-			Name:      authorinoClusterRoleName,
-			Namespace: r.Namespace,
+			Name: authorinoClusterRoleName,
 		},
 	}
 	err := r.Get(context.TODO(), client.ObjectKeyFromObject(authorinoClusterRole), authorinoClusterRole)
diff --git a/main.go b/main.go
index a8ce899..589b636 100644
--- a/main.go
+++ b/main.go
@@ -65,31 +65,29 @@ func main() {
 
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 
-	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+	managerOptions := ctrl.Options{
 		Scheme:                 scheme,
 		MetricsBindAddress:     metricsAddr,
 		Port:                   9443,
 		HealthProbeBindAddress: probeAddr,
 		LeaderElection:         enableLeaderElection,
 		LeaderElectionID:       "aac3a15d.authorino.kuadrant.net",
-	})
+	}
+
+	if ns, found := os.LookupEnv("WATCH_NAMESPACE"); found {
+		managerOptions.Namespace = ns
+	}
+
+	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), managerOptions)
 	if err != nil {
 		setupLog.Error(err, "unable to start manager")
 		os.Exit(1)
 	}
 
-	//get operator namespace
-	namespace := authorinooperatorv1beta1.AuthorinoOperatorNamespace
-	ns, found := os.LookupEnv("WATCH_NAMESPACE")
-	if found {
-		namespace = ns
-	}
-
 	if err = (&controllers.AuthorinoReconciler{
-		Client:    mgr.GetClient(),
-		Log:       ctrl.Log.WithName("authorino-operator").WithName("controller").WithName("Authorino"),
-		Scheme:    mgr.GetScheme(),
-		Namespace: namespace,
+		Client: mgr.GetClient(),
+		Log:    ctrl.Log.WithName("authorino-operator").WithName("controller").WithName("Authorino"),
+		Scheme: mgr.GetScheme(),
 	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "Authorino")
 		os.Exit(1)
```

api/v1beta1/authorino_types.go
```diff
@@ -0,0 +1,140 @@
...
+	CertPath    string `json:"certPath,omitempty"`
+	CertKeyPath string `json:"certKeyPath,omitempty"`
+}
+
+// AuthorinoStatus defines the observed state of Authorino
+type AuthorinoStatus struct {
+	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+
+	// Number of authorino instances in the cluster
```
eguzki, line 101: the comment and the type of the object does not match.

api/v1beta1/authorino_types.go
```diff
@@ -0,0 +1,140 @@
...
+	LastError string `json:"lastError"`
+
+	// Conditions is an array of the current Authorino's CR conditions
+	// Supported condition types: ConditionReady
+	// +optional
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	// +listType=map
+	// +listMapKey=type
+	Conditions []Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
```
eguzki, line 114: nice :clap:

config/authorino/crd/authorino.3scale.net_authconfigs.yaml
```diff
@@ -0,0 +1,1040 @@
+
```
eguzki, line 1: 3scale.net?

config/authorino/crd/patches/oneof_in_authconfigs.yaml
```diff
@@ -0,0 +1,57 @@
+# Enables oneOf validation for the identity/metadata/authorization modes
+
+- op: add
+  path: /spec/versions/0/schema/openAPIV3Schema/properties/spec/properties/identity/items/oneOf
```
eguzki, line 4: only one identity can be specified for each authconfig object? authorino supports many at the same time.

config/crd/bases/operator.authorino.kuadrant.net_authorinoes.yaml
```diff
@@ -0,0 +1,133 @@
+
```
eguzki, line 1: authorinoEs?

controllers/authorino_controller.go
```diff
@@ -0,0 +1,619 @@
...
+			changed = true
+		}
+	}
+	return changed
+}
+
+func (r *AuthorinoReconciler) createAuthorinoPermission(authorino *api.Authorino, operatorNamespace string) error {
+	prefix := authorino.GetName()
+	authorinoInstanceNamespace := authorino.GetNamespace()
+	authorinoClusterRoleName := "authorino-operator-authorino-manager-role"
```
guicassolato, line 328: To make it work while trying locally (following the instructions in the description of the PR), I had to change this to:

```suggestion
	authorinoClusterRoleName := "authorino-manager-role"
```

config/manager/controller_manager_config.yaml
```diff
@@ -8,4 +8,4 @@ webhook:
   port: 9443
 leaderElection:
   leaderElect: true
-  resourceName: f0fb677d.authorino.3scale.net
+  resourceName: aac3a15d.authorino.kuadrant.net
```
eguzki, line 11: kuadrant.io?

controllers/authorino_controller.go
```diff
@@ -0,0 +1,619 @@
...
+
+	return dep
+}
+
+func (r *AuthorinoReconciler) buildAuthorinoEnv(authorino *api.Authorino) []k8score.EnvVar {
+	envVar := []k8score.EnvVar{}
+
+	if !authorino.Spec.ClusterWide {
+		envVar = append(envVar, k8score.EnvVar{
+			Name:  "WATCH_NAMESPACE",
```
guicassolato, line 233: The names of other Authorino env vars seem to be defined all in constants at the level of the API definition, which is nice, but not this one.

controllers/authorino_controller.go
```diff
@@ -0,0 +1,619 @@
...
+				},
+				Spec: k8score.PodSpec{
+					ServiceAccountName: prefix + "-authorino",
+					Containers: []k8score.Container{
+						{
+							Image:           authorino.Spec.Image,
+							ImagePullPolicy: k8score.PullPolicy(authorino.Spec.ImagePullPolicy),
+							Name:            api.AuthorinoContainerName,
+							Env:             r.buildAuthorinoEnv(authorino),
+						},
```
guicassolato, line 216: For a future iteration, let's not forget to add here the `kube-rbac-proxy` sidecar that protects the `/metrics` endpoint.

eguzki (changes requested):
overall looks good. IMO the authorino_controller needs a heavy refactor to reduce complexity and DRY, but good enough for the first shoot.

config/authorino/crd/authorino.3scale.net_authconfigs.yaml
eguzki, line 1: I would automate the generation of this external resource. 

Check kuadrantctl how does this with makefile targets: https://github.com/Kuadrant/kuadrantctl/blob/ae6ee5f2c87bd10ec2053ad3b41826064d96e415/Makefile#L117
The makefile only needs one tag, and it will generate the resources in the repo.

Also a test to make sure nobody updates the auto-generated resources.

config/authorino/rbac/role.yaml
```diff
@@ -0,0 +1,45 @@
...
+  - list
+  - update
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - get
+  - list
+  - watch
```
guicassolato, line 45: We should change in Authorino first, but cannot forget reflecting the change here as well... We're missing a couple rules as part of this role:

```yaml
- apiGroups: ["authentication.k8s.io"]
  resources:
  - tokenreviews
  verbs: ["create"]
- apiGroups: ["authorization.k8s.io"]
  resources:
  - subjectaccessreviews
  verbs: ["create"]
```

These rules have never been missed really only because we are always deploying the `kube-rbac-proxy` sidecar and with that all the RBAC that provides Authorino's SA with exactly the same permission. However, one thing should not rely on the other – `kube-rbac-proxy` creates `tokenreviews` and `subjectaccessreviews` to protect the `/metrics` endpoint; Authorino creates `tokenreviews` and `subjectaccessreviews` as part of the implementation of the [Kubernetes authn](https://github.com/Kuadrant/authorino/blob/main/docs/architecture.md#kubernetes-authentication) and [Kubernetes authz](https://github.com/Kuadrant/authorino/blob/main/docs/architecture.md#kubernetes-authorization) features.

api/v1beta1/authorino_types.go
```diff
@@ -0,0 +1,140 @@
...
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+
+	Image           string               `json:"image,omitempty"`
+	Replicas        *int32               `json:"replicas,omitempty"`
+	ImagePullPolicy string               `json:"imagePullPolicy,omitempty"`
+	ClusterWide     ClusterWideAuthorino `json:"clusterWide,omitempty"`
+	Listener        Listener             `json:"listener,omitempty"`
+	OIDCServer      OIDCServer           `json:"oidcServer,omitempty"`
+	// TODO: remove after Authorino v0.5.0
```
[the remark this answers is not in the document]
guicassolato: Authorino v0.5.0 does not; v0.6.0 and on will. What I meant to be removed is the just old name of the env var, while keeping the new name.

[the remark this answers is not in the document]
guicassolato:
```suggestion
```

config/manager/controller_manager_config.yaml
guicassolato: Also mentioned in https://github.com/Kuadrant/authorino-operator/pull/1#discussion_r741964144

config/crd/bases/operator.authorino.kuadrant.net_authorinoes.yaml
guicassolato: We should change to "authorinos" IMO. https://github.com/Kuadrant/authorino-operator/pull/1#discussion_r741065960

config/authorino/crd/patches/oneof_in_authconfigs.yaml
guicassolato: Not quite. It is actually defined in https://github.com/Kuadrant/authorino/blob/main/install/crd/patches/oneof_in_authconfigs.yaml. It means exactly one of the supported identity modes must be specified – either `apiKey`, or `oidc`, or `oauths`, or `kubernetes`.

controllers/authorino_controller.go
```diff
@@ -0,0 +1,619 @@
...
+	// OIDC service
+	if authorino.Spec.OIDCServer.Port != nil {
+		envVar = append(envVar, k8score.EnvVar{
+			Name:  api.OIDCHTTPPort,
+			Value: fmt.Sprint(authorino.Spec.OIDCServer.Port),
+		})
+	}
+	if authorino.Spec.OIDCServer.CertKeyPath != "" {
+		envVar = append(envVar, k8score.EnvVar{
+			Name:  api.OIDCTSLCertPath,
```
guicassolato, line 281:
```suggestion
			Name:  api.OIDCTLSCertPath,
```

controllers/authorino_controller.go
```diff
@@ -0,0 +1,619 @@
...
+	}
+	if authorino.Spec.Listener.CertPath != "" {
+		envVar = append(envVar, k8score.EnvVar{
+			Name:  api.TSLCertPath,
+			Value: authorino.Spec.Listener.CertPath,
+		})
+	}
+	if authorino.Spec.Listener.CertKeyPath != "" {
+		envVar = append(envVar, k8score.EnvVar{
+			Name:  api.TSLCertKeyPath,
```
guicassolato, line 267:
```suggestion
			Name:  api.TLSCertKeyPath,
```

controllers/authorino_controller.go
```diff
@@ -0,0 +1,619 @@
...
+	// external auth service via GRPC
+	if authorino.Spec.Listener.Port != nil {
+		envVar = append(envVar, k8score.EnvVar{
+			Name:  api.ExtAuthGRPCPort,
+			Value: fmt.Sprint(authorino.Spec.Listener.Port),
+		})
+	}
+	if authorino.Spec.Listener.CertPath != "" {
+		envVar = append(envVar, k8score.EnvVar{
+			Name:  api.TSLCertPath,
```
guicassolato, line 261:
```suggestion
			Name:  api.TLSCertPath,
```

api/v1beta1/authorino_types.go
```diff
@@ -0,0 +1,140 @@
...
+	AuthConfigLabelSelectors string `json:"authConfigLabelSelectors,omitempty"`
+	SecretLabelSelectors     string `json:"secretLabelSelectors,omitempty"`
+}
+
+type ClusterWideAuthorino bool
+
+type Listener struct {
+	Port        *int32 `json:"port,omitempty"`
+	Tls         bool   `json:"tsl,omitempty"`
+	CertPath    string `json:"certPath,omitempty"`
```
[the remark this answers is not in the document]
guicassolato: It's an Authorino thing, but we're missing a few bits here. In Authorino, these fields are supposed to store the paths to the cert and cert key files to be mounted in the Authorino container out of keys of a `Secret` that is generated by cert-manager.

At the level of the operator, we should probably make these references to the `Secret` and abstract the details of where in file system the keys are mounted as files.

Same for the OIDC TLS cert.

Makefile
```diff
@@ -120,15 +119,21 @@ deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in
 undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/config.
 	$(KUSTOMIZE) build config/default | kubectl delete -f -
 
+install-authorino: ## install RBAC and CRD for authorino
```
eguzki, line 122: dependency to generate `kustomize` tool? I just run `make install-authorino` and got the error `bin/kustomize: No such file or directory`

config/authorino/crd/patches/oneof_in_authconfigs.yaml
eguzki: ok, I see. It is not about the `spec.identity` array, it is about the `items` of it. They must comply with one and only one of the specs there. 

The validation accepts empty array:
```yaml
---                                      
apiVersion: authorino.3scale.net/v1beta1 
kind: AuthConfig                         
metadata:                                
  name: my-api-protection                
spec:                                    
  hosts:                                 
  ⦙ - my-api.io                          
  identity: []                           

```

guicassolato: Right now empty array means always 401. Hopefully #112 will fix that.

api/v1beta1/authorino_types.go
jjaferson: https://github.com/Kuadrant/authorino-operator/issues/2

Makefile
jjaferson: Yup, we can create an issue that remove the Authorino's CRDs from the cluster

controllers/authorino_controller.go
jjaferson: I didn't include this to the list of env vars defined by Authorino because `WATCH_NAMESPACE` is defined by the `operator sdk`, therefore, authorino doesn't have much control over it but no harm in adding this env var to the list

jjaferson:
> overall looks good. IMO the authorino_controller needs a heavy refactor to reduce complexity and DRY, but good enough for the first shoot.

Thanks for reviewing @eguzki I agree, we can make the optimize the code and improve readability and reusability. I will start working on it once we get all the initial steps in place :)

main.go
jjaferson: > My understanding for "`WATCH_NAMESPACE`" is that this should be the namespace the operator will be watching for `Authorino` CRs (if not the entire cluster). Instead, it here seems to be used as the namespace where the operator is running maybe?
> 
> We here seems to be looking for a `ClusterRole` in the namespace that matches the value of this variable, with default to "`authorino-operator`", which doesn't sound right to me. We probably don't even need the name of the namespace to get the `ClusterRole`.
> 
> Maybe rename this variable... and either use `WATCH_NAMESPACE` as expected (an option of the manager, not of the controller) or make the operator always cluster-scoped (i.e. get rid of `WATCH_NAMESPACE`).

Spot on, we don't need the namespace to fetch the cluster role and as the operator is cluster-scoped we might not need this variable at all

api/v1beta1/authorino_types.go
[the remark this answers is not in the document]
jjaferson: @guicassolato so if it's a reference to a secret, it should only have the secret name or should the operator create the secret with a placeholder for the certs?

I'll also create an issue to make the operator install the cert-manager, so the operator could handle the creation of the secret one the cert-manager is installed and running?

config/crd/bases/operator.authorino.kuadrant.net_authorinoes.yaml
jjaferson: @eguzki this name structure is generated by the operator-sdk, do you know if we can customize it?

eguzki: `operator-sdk create api` has this `--plural`. Did you tried it?

config/authorino/rbac/role.yaml
jjaferson: Right, I added this permission to the ClusterRole.

@guicassolato When the operator is ready for the first release should we remove those RBAC rules from the authorino repo and have it only on the repo of the operator?

jjaferson: > Right, I added this permission to the ClusterRole.
> 
> @guicassolato When the operator is ready for the first release should we remove those RBAC rules from the authorino repo and have it only on the repo of the operator?

Actually from @eguzki [comment](https://github.com/Kuadrant/authorino-operator/pull/1#discussion_r742068502), we can generate it automatically from authorino's repo so there is no need for the operator to host the CRDs and permissions.

config/crd/bases/operator.authorino.kuadrant.net_authorinoes.yaml
jjaferson: I got it working by editing the `//+kubebuilder:resource:path="authorinos"` to `authorino_types.go` https://book.kubebuilder.io/reference/markers/crd.html?highlight=crd,plural#crd-generation

guicassolato:
Tested on a local cluster (without TLS):

```shell
kind create cluster --name authorino
make install

watch -n 3 "kubectl -n authorino-operator get pods | grep -viE 'Completed|OOMKilled'"

make docker-build
kind load docker-image authorino-operator:latest --name authorino
make deploy
kubectl -n authorino-operator patch deployment/authorino-operator-controller-manager -p '{"spec":{"template":{"spec":{"containers":[{"name":"manager","imagePullPolicy":"IfNotPresent"}]}}}}'

kubectl create namespace myapp
watch -n 3 "kubectl -n myapp get pods | grep -viE 'Completed|OOMKilled'"

kubectl -n myapp apply -f https://raw.githubusercontent.com/Kuadrant/authorino/main/examples/talker-api/talker-api-deploy.yaml

kubectl -n myapp apply -f -<<EOF
kind: Authorino
metadata:
  name: authorino
  namespace: myapp
spec:
  image: quay.io/3scale/authorino:latest
  replicas: 1
  clusterWide: false
EOF

curl -L https://raw.githubusercontent.com/Kuadrant/authorino/main/examples/envoy/overlays/notls/configmap.yaml | sed -E 's/authorino-authorization/authorino-authorino-authorization/g' | kubectl -n myapp apply -f -
kubectl -n myapp apply -f https://raw.githubusercontent.com/Kuadrant/authorino/main/examples/envoy/base/envoy.yaml

kubectl -n myapp apply -f -<<EOF
apiVersion: authorino.3scale.net/v1beta1
kind: AuthConfig
metadata:
  name: talker-api-protection
spec:
  hosts:
    - talker-api
  identity:
    - name: friends
      apiKey:
        labelSelectors:
          group: friends
      credentials:
        in: authorization_header
        keySelector: APIKEY
EOF

kubectl -n myapp apply -f -<<EOF
apiVersion: v1
kind: Secret
metadata:
  name: friend-1-api-key-1
  labels:
    authorino.3scale.net/managed-by: authorino
    group: friends
stringData:
  api_key: ndyBzreUzF4zqDQsqSPMHkRhriEOtcRx
type: Opaque
EOF

kubectl -n myapp port-forward deployment/envoy 8000:8000
curl -H 'Host: talker-api' -H 'Authorization: APIKEY ndyBzreUzF4zqDQsqSPMHkRhriEOtcRx' http://localhost:8000/hello
```

We still need to fix the TLS cert fields, changing to `secretRef` probably, but so far so good.

api/v1beta1/authorino_types.go
```diff
@@ -0,0 +1,146 @@
...
+	ExtAuthGRPCPort          string = "EXT_AUTH_GRPC_PORT"
+	EnvVarTlsCert            string = "TLS_CERT"
+	EnvVarTlsCertKey         string = "TLS_CERT_KEY"
+	OIDCHTTPPort             string = "OIDC_HTTP_PORT"
+	EnvVarOidcTlsCertPath    string = "OIDC_TLS_CERT"
+	EnvVarOidcTlsCertKeyPath string = "OIDC_TLS_CERT_KEY"
+	AuthConfigLabelSelector  string = "AUTH_CONFIG_LABEL_SELECTOR"
+	SecretLabelSelector      string = "SECRET_LABEL_SELECTOR"
+
+	//
```
guicassolato, line 45:
```suggestion
	// Authorino TLS file paths
```

?

config/authorino/rbac/authconfig_viewer_role.yaml
```diff
@@ -0,0 +1,20 @@
+# permissions for end users to view services.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: authconfig-viewer-role
```
guicassolato, line 5: We need to do this change in http://github.com/kuadrant/authorino – i.e. fix the name of the file and name of the resource.

Should we be committing these `config/authorino/rbac/*` files if Authorino is installed by git cloning the repo and installing from the manifests hosted there?

controllers/authorino_controller.go
```diff
@@ -0,0 +1,703 @@
...
+				},
+				Spec: k8score.PodSpec{
+					ServiceAccountName: prefix + "-authorino",
+				},
+			},
+		},
+	}
+
+	authorinoContainer := k8score.Container{
+		Image:           authorino.Spec.Image,
```
guicassolato, line 234: Should we provide a default image?

Makefile
```diff
@@ -120,15 +119,25 @@ deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in
 undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/config.
 	$(KUSTOMIZE) build config/default | kubectl delete -f -
 
+AUTHORINO_VERSION=v0.4.0
```
guicassolato, line 122:
```suggestion
AUTHORINO_VERSION=v0.5.0
```

or

```suggestion
AUTHORINO_VERSION=main
```

controllers/authorino_controller.go
```diff
@@ -0,0 +1,703 @@
...
+		ImagePullPolicy: k8score.PullPolicy(authorino.Spec.ImagePullPolicy),
+		Name:            api.AuthorinoContainerName,
+		Env:             r.buildAuthorinoEnv(authorino),
+	}
+
+	if &authorino.Spec.Listener.Tls.Enabled != nil && authorino.Spec.Listener.Tls.Enabled == true {
```
guicassolato, line 240:
```suggestion
	if enabled := authorino.Spec.Listener.Tls.Enabled; &enabled == nil || enabled {
```

config/authorino/rbac/authconfig_viewer_role.yaml
guicassolato: About committing these files, having all the manifests (Operator's and Authorino's) in one single place – actually even in one single multi-document YAML file – can be useful to simplify installation. Instead of cloning the repo, one could install by applying directly sth like https://raw.githubusercontent.com/kuadrant/authorino-operator/main/config/manifests.yaml, plus deploying the operator from a pre-built public image.

controllers/authorino_controller.go
jjaferson: Good question, we could and use the image version that is compatible with the operator?

controllers/authorino_controller.go
jjaferson: Should we define a pointer for the bool or just check the variable address is nil using the `&` sign? 

```&enabled == nil || enabled```

Comment (guicassolato) [reply controllers/authorino_controller.go:240]: I think checking the memory address of the variable won't work, as boolean values are set to `false` by default, and [`&false != &nil`](https://play.golang.org/p/UvezGdQbpbK).

File: config/authorino/rbac/authconfig_viewer_role.yaml (hunk 46 above)
Comment (jjaferson) [reply config/authorino/rbac/authconfig_viewer_role.yaml:5]: Authorino's CRD and RBAC will be installed via the operator by applying the manifests available on the authorino's repo once this [issue](https://github.com/Kuadrant/authorino/issues/190) is resolved

File: controllers/authorino_controller.go (hunk 49 above)
Comment (jjaferson) [reply controllers/authorino_controller.go:240]: True

File: Makefile [hunk 50]
```diff
@@ -28,17 +28,19 @@ BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL)
...
+# authorino.kuadrant.io/authorino-operator-bundle:$VERSION and authorino.kuadrant.io/authorino-operator-catalog:$VERSION.
+IMAGE_TAG_BASE ?= authorino.kuadrant.io/authorino-operator
 
 # BUNDLE_IMG defines the image:tag used for the bundle.
 # You can use it as an arg. (E.g make bundle-build BUNDLE_IMG=<some-registry>/<project-name-bundle>:<tag>)
 BUNDLE_IMG ?= $(IMAGE_TAG_BASE)-bundle:v$(VERSION)
 
 # Image URL to use all building/pushing image targets
-IMG ?= controller:latest
+IMG ?= authorino-operator:latest
```
guicassolato, line 39: Can we please make `quay.io/3scale/authorino-operator:latest` the default here?

Plus future enhancement: CI to automatically push to quay.io when merged to main branch ([example](https://github.com/Kuadrant/authorino/blob/main/.github/workflows/build-latest.yaml)).

controllers/authorino_controller.go
jjaferson: For now I set the `latest` tag as the default version but we can change this when there is a version that is not compatible with the operator

api/v1beta1/authorino_types.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+type OIDCServer struct {
+	Port *int32 `json:"port,omitempty"`
+	Tls  Tls    `json:"tls,omitempty"`
+}
+
+type Tls struct {
+	Enabled        *bool  `json:"enabled,omitempty"`
+	CertSecretName string `json:"certSecretName"`
```
guicassolato, line 100:
```suggestion
	CertSecretName *k8score.LocalObjectReference `json:"certSecretName,omitempty"`
```

config/samples/authorino-operator_v1beta1_authorino.yaml
```diff
@@ -0,0 +1,20 @@
...
+spec:
+  image: quay.io/3scale/authorino:latest
+  replicas: 1
+  imagePullPolicy: Always
+  clusterWide: true
+  listener:
+    port: 
+    tls:
+      enabled: true # default
+      certSecretRef: authorino-cert # secret must contain `tls.crt` and `tls.key` entries
```
guicassolato, line 15:
```suggestion
      certSecretRef:
        name: authorino-cert # secret must contain `tls.crt` and `tls.key` entries
```

config/samples/authorino-operator_v1beta1_authorino.yaml
```diff
@@ -0,0 +1,20 @@
...
+  oidcServer:
+    port:
+    tls:
+      enabled: true # default
+      certSecretRef: authorino-cert # secret must contain `tls.crt` and `tls.key` entries
```
guicassolato, line 20:
```suggestion
      certSecretRef:
        name: authorino-cert # secret must contain `tls.crt` and `tls.key` entries
```

config/rbac/authorino_editor_role.yaml
```diff
@@ -0,0 +1,24 @@
+# permissions for end users to edit authorinos.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: authorino-editor-role
```
guicassolato, line 5: This will collide with the name of Authorino's `editor-role` after https://github.com/Kuadrant/authorino/pull/191 was merged. See https://github.com/Kuadrant/authorino/blob/05f151cfc1bcdf4bcd4cfc95efc367ead80df1dc/install/manifests.yaml#L1110

Maybe we should fix it in https://github.com/kuadrant/authorino, i.e. change the name of the role there to `authconfig-editor-role` (`authorino-authconfig-editor-role` after prefixed)

config/rbac/authorino_viewer_role.yaml
```diff
@@ -0,0 +1,20 @@
+# permissions for end users to view authorinos.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: authorino-viewer-role
```
guicassolato, line 5: Same as https://github.com/Kuadrant/authorino-operator/pull/1/files#r750634235

config/samples/authorino-operator_v1beta1_authorino.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: operator.authorino.kuadrant.io/v1beta1
+kind: Authorino
+metadata:
+  name: authorino-sample22
```
guicassolato, line 4: Numerology? 😜

guicassolato (approved):
Tested with the following script:

```sh
kind create cluster --name authorino

# operator
make install deploy OPERATOR_IMAGE=authorino-operator:local

kubectl create namespace myapp

# upstream
kubectl -n myapp apply -f https://raw.githubusercontent.com/Kuadrant/authorino/main/examples/talker-api/talker-api-deploy.yaml

# authorino (without tls)
kubectl -n myapp apply -f -<<EOF
kind: Authorino
metadata:
  name: authorino
spec:
  replicas: 1
  clusterWide: false
  listener:
    tls:
      enabled: false
  oidcServer:
    tls:
      enabled: false
EOF

# envoy (without tls)
curl -L https://raw.githubusercontent.com/Kuadrant/authorino/main/examples/envoy/overlays/notls/configmap.yaml | sed -E 's/authorino-authorization/authorino-authorino-authorization/g' | kubectl -n myapp apply -f -
kubectl -n myapp apply -f https://raw.githubusercontent.com/Kuadrant/authorino/main/examples/envoy/base/envoy.yaml

# authconfig
kubectl -n myapp apply -f -<<EOF
apiVersion: authorino.3scale.net/v1beta1
kind: AuthConfig
metadata:
  name: talker-api-protection
spec:
  hosts:
    - talker-api
  identity:
    - name: friends
      apiKey:
        labelSelectors:
          group: friends
      credentials:
        in: authorization_header
        keySelector: APIKEY
EOF

# consume
kubectl -n myapp apply -f -<<EOF
apiVersion: v1
kind: Secret
metadata:
  name: friend-1-api-key-1
  labels:
    authorino.3scale.net/managed-by: authorino
    group: friends
stringData:
  api_key: ndyBzreUzF4zqDQsqSPMHkRhriEOtcRx
type: Opaque
EOF

kubectl -n myapp port-forward deployment/envoy 8000:8000 &
curl -H 'Host: talker-api' -H 'Authorization: APIKEY ndyBzreUzF4zqDQsqSPMHkRhriEOtcRx' http://localhost:8000/hello

# tls certs
CURRENT_DIR=$PWD; AUTHORINO_DIR=$(mktemp -d); cd $AUTHORINO_DIR
git clone --depth 1 --branch main https://github.com/kuadrant/authorino.git && cd authorino
make cert-manager
make certs AUTHORINO_NAMESPACE=myapp
rm -rf $AUTHORINO_DIR; cd $CURRENT_DIR

# authorino (with tls)
kubectl -n myapp apply -f -<<EOF
kind: Authorino
metadata:
  name: authorino
spec:
  replicas: 1
  clusterWide: false
  listener:
    tls:
      certSecretRef:
        name: authorino-server-cert
  oidcServer:
    tls:
      certSecretRef:
        name: authorino-oidc-server-cert
EOF

# envoy (with tls)
curl -L https://raw.githubusercontent.com/Kuadrant/authorino/main/examples/envoy/overlays/tls/configmap.yaml | sed -E 's/authorino-authorization/authorino-authorino-authorization/g' | kubectl -n myapp apply -f -
kubectl -n myapp apply -f https://raw.githubusercontent.com/Kuadrant/authorino/main/examples/envoy/base/envoy.yaml
kubectl -n myapp patch deployment/envoy -p '{"spec":{"template":{"spec":{"volumes":[{"name":"authorino-ca-cert","secret":{"defaultMode":420,"secretName":"authorino-ca-cert"}}],"containers":[{"name":"envoy","volumeMounts":[{"name":"authorino-ca-cert","subPath":"ca.crt","mountPath":"/etc/ssl/certs/authorino-ca-cert.crt","readOnly":true}]}]}}}}'

# consume
curl -H 'Host: talker-api' -H 'Authorization: APIKEY ndyBzreUzF4zqDQsqSPMHkRhriEOtcRx' http://localhost:8000/hello

# cleanup
kind delete cluster --name authorino
```

Works perfectly! Congratz and thank you, @jjaferson!

config/rbac/authorino_editor_role.yaml
guicassolato: Done in https://github.com/Kuadrant/authorino/pull/192.

config/rbac/authorino_viewer_role.yaml
guicassolato: Done in https://github.com/Kuadrant/authorino/pull/192.

guicassolato (approved):

eguzki (approved):
Good job :medal_military:

Issue #12, opened 2021-11-29
Authorino CRD status update


jjaferson:
Implement a logic to update the Authorino CRs status block based on the [conditions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties) specification.

PR #13, opened 2021-12-06
Adds status conditions to the Authorino API definition

jjaferson:
This PR adds [conditions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties) to the status block of the authorino API definition, this will provide a higher-level status reporting  for the authorino instances managed by the operator.

The PR also contains some code refactoring to the controller.

Closes https://github.com/Kuadrant/authorino-operator/issues/12

### Verification steps
1) Install and deploy the operator from this branch 
2) Create a simple instance of authorino
```bash
kubectl -n myapp apply -f -<<EOF
kind: Authorino
metadata:
  name: authorino
spec:
  replicas: 1
  clusterWide: false
  listener:
    tls:
      enabled: false
  oidcServer:
    tls:
      enabled: false
EOF
```
3) Verify conditions were set to the status definition of the API
```yaml
status:
  conditions:
  - lastTransitionTime: "2021-12-06T14:19:12Z"
    reason: Provisioned
    status: "True"
    type: Ready
```


config/crd/bases/operator.authorino.kuadrant.io_authorinos.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.6.1
+    controller-gen.kubebuilder.io/version: v0.4.1
```
guicassolato, line 7: The Makefile says v0.6.1. This change is probably due to an older version installed locally.

https://github.com/Kuadrant/authorino-operator/blob/c5c0510c810f94a8267da238705799ea1af32b6e/Makefile#L134

guicassolato (approved):

guicassolato (approved):

PR #20, opened 2021-12-19
Option to mount extra volumes in the Authorino containers

guicassolato:
- [x] Add `spec.volumes` ~and `spec.volumeMounts`~ to the API →  optional arrays of the `Authorino` CR whose ~elements are of the core Kubernetes types [`Volume`](https://pkg.go.dev/k8s.io/api@v0.21.2/core/v1#Volume) and [`VolumeMount`](https://pkg.go.dev/k8s.io/api@v0.21.2/core/v1#VolumeMount) respectively~ `Items` field is an array of specs to define both deployment `volumes` and container `volumeMounts`.
- [x] Bugfix: volumeMount config not being pushed to the authorino container template →  causes enabling TLS in the authorino instances to fail, apart from the fix being a requirement for this change to work.

For the comparison between desired and current states, the reconciler checks only for the number of items in the `volume` and `volumeMount` arrays and the names of the objects in those arrays. This is relatively limited as it will cause the operator to miss changes in the actual specs of those volume and volumeMounts – e.g. different name of Secret or ConfigMap volume source referred, different list of items from a volume source to be mounted as file, etc. Nevertheless, this is more reliable than using `reflect.DeepEqual` or comparing the marshalled byte arrays of the resources, which could cause an endless loop of states constantly being considered different and the status of the resource never to be updated to the final status "Provisioned".

----

**Verification steps**

```sh
cd $GOPATH/src/github.com/kuadrant/authorino
git checkout main
make cluster
sed -i 's/FROM gcr\.io\/distroless\/static:nonroot/FROM gcr\.io\/distroless\/static:debug/' Dockerfile
make local-build
git checkout
make cert-manager
make install-operator OPERATOR_VERSION=volumes

cd $GOPATH/src/github.com/kuadrant/authorino-operator
git checkout volumes
kubectl -n authorino-operator patch deployment/authorino-operator-controller-manager -p '{"spec":{"template":{"spec":{"containers":[{"name":"manager","image":"authorino-operator:local"}]}}}}'

cd $GOPATH/src/github.com/kuadrant/authorino
make namespace
touch tmp/keycloak.crt
kubectl -n authorino create configmap keycloak-tls-cert --from-file=tmp/keycloak.crt

kubectl -n authorino apply -f -<<EOF
kind: Authorino
metadata:
  name: authorino
spec:
  image: authorino:local
  imagePullPolicy: IfNotPresent
  clusterWide: false
  volumes:
    items:
    - name: keycloak-tls-cert-ext
      mountPath: /etc/ssl/certs
      configMaps:
      - keycloak-tls-cert
      items:
      - key: keycloak.crt
        path: keycloak.crt
    defaultMode: 420
  listener:
    tls:
      enabled: false
  oidcServer:
    tls:
      enabled: false
EOF

kubectl -n authorino exec $(kubectl -n authorino get pods -l authorino-resource=authorino -o name) -- ls /etc/ssl/certs
# keycloak.crt
```


guicassolato:
Leaving here for reference the approach used by the [Keycloak Operator,](https://github.com/keycloak/keycloak-operator) where only `ConfigMap` and `Secret` volume sources are supported, and the specs needed for `volumes` and `volumeMounts` are mixed into one same type in the CRD:

- https://github.com/keycloak/keycloak-operator/blob/main/deploy/crds/keycloak.org_keycloaks_crd.yaml#L883
- https://github.com/keycloak/keycloak-operator/blob/dc30f6efbcffb9ff8a6c49585ea9e19793fb21ac/pkg/apis/keycloak/v1alpha1/keycloak_types.go#L140

jjaferson (approved):
Changes look good, thanks @guicassolato

PR #21, opened 2022-01-10
Authorino K8s auth ClusterRole

guicassolato:
Add support for `authorino-manager-k8s-auth-role` `Role`.

Related to https://github.com/Kuadrant/authorino/pull/211

Plus:
- Fixes owner references of Authorino `Service`, `ServiceAccount` and other resources.
- Fixes `ClusterRoleBinding`s created to append new subjects instead of always creating another binding – prevents overriding the cluster-scoped resource when there are more than one instance of Authorino with the same name (in different k8s namespaces of course)
- ~Prepends the name of the `Authorino` CR's namespace to the name of the `ClusterRoleBinding`s created.~

TODO:
- [ ] ~Fix unique names of the `ClusterRoleBinding`s – `<cr-namespace>-<cr-name>-<crb-name>` is not enough to prevent clashes (e.g. `namespace: foo-x / name: bar` and `namespace: foo / name: x-bar`)~
- [x] Fix cleanup on deletion of `Authorino` custom resources – Due to the fix for appending subjects to the `ClusterRoleBinding` instead of creating new ones, the cleanup is now broken. When the `Authorino` resource is deleted, the corresponding `ServiceAccount` is not removed from the list of subjects of the `ClusterRoleBinding`s. To fix that, we can either
    - ~get back to creating one set of `ClusterRoleBinding`s per Authorino instance (adding the name and the namespace of the `Authorino` resource to the name of the `ClusterRoleBinding`, to avoid overwriting)~ – it will not work due to k8s ownerreferences not allowed between cluster-scoped dependents and namespaced owners ([ref](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/)), or
    - implement the removal of the subject from the `ClusterRoleBinding`s on 404 Not Found of `Authorino` CRs caught by the reconciler. [DONE]

### Verification steps

Build and deploy images of Authorino and Authorino Operator that contains the required changes:

```sh
cd $GOPATH/src/github.com/kuadrant/authorino-operator
git checkout authorino-k8s-auth-role

cd $GOPATH/src/github.com/kuadrant/authorino
git checkout fix/k8s-auth-namespaced
make cluster
make cert-manager
make local-build # kind load docker-image authorino:local --name authorino
make install-operator OPERATOR_VERSION=authorino-k8s-auth-role
kubectl -n authorino-operator patch deployment/authorino-operator-controller-manager -p '{"spec":{"template":{"spec":{"containers":[{"name":"manager","image":"authorino-operator:local"}]}}}}'
make install
make namespace
make deploy

kubectl -n authorino apply -f -<<EOF
kind: Authorino
metadata:
  name: authorino-2
spec:
  clusterWide: false
  listener:
    tls:
      enabled: false
  oidcServer:
    tls:
      enabled: false
EOF
```

Check the resources created:

```sh
kubectl get clusterroles -o name | grep authorino-manager
# clusterrole.rbac.authorization.k8s.io/authorino-manager-k8s-auth-role
# clusterrole.rbac.authorization.k8s.io/authorino-manager-role

kubectl get clusterrolebindings | grep authorino-manager
# authorino-k8s-auth                                     ClusterRole/authorino-manager-k8s-auth-role                                        46s

kubectl get clusterrolebindings/authorino-k8s-auth -o yaml
# apiVersion: rbac.authorization.k8s.io/v1
# kind: ClusterRoleBinding
# metadata:
#   creationTimestamp: "2022-01-12T12:08:58Z"
#   name: authorino-k8s-auth
#   resourceVersion: "3734"
#   uid: 3c99dc92-2173-4c4c-bec5-0d94906dd8d7
# roleRef:
#   apiGroup: rbac.authorization.k8s.io
#   kind: ClusterRole
#   name: authorino-manager-k8s-auth-role
# subjects:
# - kind: ServiceAccount
#   name: authorino-authorino
#   namespace: authorino
# - kind: ServiceAccount
#   name: authorino-2-authorino
#   namespace: authorino

kubectl -n authorino get serviceaccounts -o name | grep authorino
# serviceaccount/authorino-2-authorino
# serviceaccount/authorino-authorino

kubectl -n authorino get rolebindings | grep authorino
# authorino-2-authorino                   ClusterRole/authorino-manager-role    106s
# authorino-2-authorino-leader-election   Role/authorino-leader-election-role   106s
# authorino-authorino                     ClusterRole/authorino-manager-role    3m49s
# authorino-authorino-leader-election     Role/authorino-leader-election-role   3m49s
```

Delete one of the Authorino instances and check again:

```sh
kubectl -n authorino delete authorinos/authorino
# authorino.operator.authorino.kuadrant.io "authorino" deleted

kubectl get clusterrolebindings | grep authorino-manager
# authorino-k8s-auth                                     ClusterRole/authorino-manager-k8s-auth-role                                        4m58s

kubectl get clusterrolebindings/authorino-k8s-auth -o yaml
# apiVersion: rbac.authorization.k8s.io/v1
# kind: ClusterRoleBinding
# metadata:
#   creationTimestamp: "2022-01-12T12:08:58Z"
#   name: authorino-k8s-auth
#   resourceVersion: "3734"
#   uid: 3c99dc92-2173-4c4c-bec5-0d94906dd8d7
# roleRef:
#   apiGroup: rbac.authorization.k8s.io
#   kind: ClusterRole
#   name: authorino-manager-k8s-auth-role
# subjects:
# - kind: ServiceAccount
#   name: authorino-2-authorino
#   namespace: authorino
```

### DIFF
--- a/controllers/authorino_controller.go
+++ b/controllers/authorino_controller.go
@@ -48,9 +48,12 @@ const (
 	tlsCertName     string = "tls-cert"
 	oidcTlsCertName string = "oidc-cert"
 
-	authorinoClusterRoleName        string = "authorino-manager-role"
-	authorinoK8sAuthClusterRoleName string = "authorino-manager-k8s-auth-role"
-	leaderElectionRoleName          string = "authorino-leader-election-role"
+	authorinoClusterRoleName                     string = "authorino-manager-role"
+	authorinoK8sAuthClusterRoleName              string = "authorino-manager-k8s-auth-role"
+	leaderElectionRoleName                       string = "authorino-leader-election-role"
+	authorinoManagerClusterRoleBindingNameSuffix string = "authorino"
+	authorinoK8sAuthClusterRoleBindingNameSuffix string = "authorino-k8s-auth"
+	authorinoLeaderElectionRoleBindingNameSuffix string = "authorino-leader-election"
 )
 
 //+kubebuilder:rbac:groups=operator.authorino.kuadrant.io,resources=authorinos,verbs=get;list;watch;create;update;patch;delete
@@ -162,6 +165,7 @@ func (r *AuthorinoReconciler) getAuthorinoInstance(namespacedName types.Namespac
 	if err != nil {
 		if errors.IsNotFound(err) {
 			r.Log.Info("Authorino CR not found.")
+			r.cleanupClusterScopedPermissions(context.Background(), namespacedName)
 			return nil, nil
 		}
 		return nil, err
@@ -425,9 +429,39 @@ func (r *AuthorinoReconciler) authorinoDeploymentChanges(existingDeployment, des
 	return false
 }
 
+func authorinoServiceAccountName(crName string) string {
+	return fmt.Sprintf("%s-authorino", crName)
+}
+
+func (r *AuthorinoReconciler) createAuthorinoServiceAccount(authorino *api.Authorino) (*k8score.ServiceAccount, error) {
+	var logger = r.Log
+	saName := authorinoServiceAccountName(authorino.Name)
+	saNamespace := authorino.Namespace
+	sa := &k8score.ServiceAccount{}
+	if err := r.Get(context.TODO(), namespacedName(saNamespace, saName), sa); err != nil {
+		if errors.IsNotFound(err) {
+			// ServiceAccount doesn't exit - create one
+			sa.Name = saName
+			sa.Namespace = saNamespace
+			_ = ctrl.SetControllerReference(authorino, sa, r.Scheme)
+			if err := r.Client.Create(context.TODO(), sa); err != nil {
+				return nil, r.wrapErrorWithStatusUpdate(
+					logger, authorino, r.setStatusFailed(api.AuthorinoUnableToCreateServiceAccount),
+					fmt.Errorf("failed to create %s ServiceAccount, err: %v", saName, err),
+				)
+			}
+		}
+		return nil, r.wrapErrorWithStatusUpdate(
+			logger, authorino, r.setStatusFailed(api.AuthorinoUnableToGetServiceAccount),
+			fmt.Errorf("failed to get %s ServiceAccount, err: %v", saName, err),
+		)
+	} else {
+		return sa, nil
+	}
+}
+
 func (r *AuthorinoReconciler) createAuthorinoPermission(authorino *api.Authorino, operatorNamespace string) error {
 	var logger = r.Log
-	resourcePrefixName := authorino.Name
 
 	clNsdName := namespacedName(authorino.Namespace, authorinoClusterRoleName)
 	authorinoClusterRole := &k8srbac.ClusterRole{}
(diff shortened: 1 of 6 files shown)


controllers/authorino_controller.go
```diff
@@ -424,9 +429,39 @@ func (r *AuthorinoReconciler) authorinoDeploymentChanges(existingDeployment, des
 	return false
 }
 
+func authorinoServiceAccountName(crName string) string {
```
jjaferson, line 432: Should we consider moving this function to the resources package under the `k8s_rbac.go` in case it needs to be used somewhere else in the code? https://github.com/Kuadrant/authorino-operator/blob/main/pkg/resources/k8s_rbac.go

controllers/authorino_controller.go
```diff
@@ -424,9 +429,39 @@ func (r *AuthorinoReconciler) authorinoDeploymentChanges(existingDeployment, des
...
+					logger, authorino, r.setStatusFailed(api.AuthorinoUnableToCreateServiceAccount),
+					fmt.Errorf("failed to create %s ServiceAccount, err: %v", saName, err),
+				)
+			}
+		}
+		return nil, r.wrapErrorWithStatusUpdate(
+			logger, authorino, r.setStatusFailed(api.AuthorinoUnableToGetServiceAccount),
+			fmt.Errorf("failed to get %s ServiceAccount, err: %v", saName, err),
+		)
+	} else {
```
jjaferson, line 458: Do we need this else statement here? We could only return the service account because if an error occur it will be returned already as you can see in line 454

controllers/authorino_controller.go
```diff
@@ -443,73 +478,96 @@ func (r *AuthorinoReconciler) createAuthorinoPermission(authorino *api.Authorino
...
-			sa.Namespace = authorino.Namespace
-			if err := r.Client.Create(context.TODO(), sa); err != nil {
-				return r.wrapErrorWithStatusUpdate(
-					logger, authorino, r.setStatusFailed(api.AuthorinoUnableToCreateServiceAccount),
-					fmt.Errorf("failed to create %s ServiceAccount, err: %v", saName, err),
-				)
-			}
+	if sa, err := r.createAuthorinoServiceAccount(authorino); err != nil {
+		return err
+	} else {
```
jjaferson, line 483: is the else statement needed here as if an error occur it will be returned before?

controllers/authorino_controller.go
```diff
@@ -443,73 +478,96 @@ func (r *AuthorinoReconciler) createAuthorinoPermission(authorino *api.Authorino
...
+}
 
-	// creates the ClusterRoleBinding/RoleBinding depending on type of installation
-	// if ClusterWide true - ClusterRoleBinding
-	// if ClusterWide false - RoleBinding
-	var binding client.Object = authorinoResources.GetAuthorinoClusterRoleBinding(authorinoClusterRoleName, sa.Name, sa.Namespace)
-	if !authorino.Spec.ClusterWide {
-		binding = authorinoResources.GetAuthorinoRoleBinding(authorinoClusterRoleName, sa.Name, sa.Namespace)
-		binding.SetNamespace(authorino.Namespace)
+func authorinoRoleBindingName(crName, roleBindingNameSuffix string, clusterScoped bool) string {
```
jjaferson, line 500: Could we move this to the resource package under the [k8s_rbac.go](https://github.com/Kuadrant/authorino-operator/blob/main/pkg/resources/k8s_rbac.go) file?

controllers/authorino_controller.go
```diff
@@ -443,73 +478,96 @@ func (r *AuthorinoReconciler) createAuthorinoPermission(authorino *api.Authorino
...
+	var roleBindingName = authorinoRoleBindingName(authorino.Name, roleBindingNameSuffix, clusterScoped)
+	var roleBindingSearchName types.NamespacedName
+
+	if clusterScoped {
+		roleBinding = authorinoResources.GetAuthorinoClusterRoleBinding(roleName, serviceAccount)
+		roleBindingSearchName = types.NamespacedName{Name: roleBindingName}
+	} else {
+		roleBinding = authorinoResources.GetAuthorinoRoleBinding(roleName, serviceAccount)
+		roleBinding.SetNamespace(authorino.Namespace)
+		roleBindingSearchName = types.NamespacedName{Name: roleBindingName, Namespace: authorino.Namespace}
```
jjaferson, line 521: We could use the [namespacedName](https://github.com/Kuadrant/authorino-operator/blob/788fd2809c021eef15f1dac6b1a2703d95ff46d2/controllers/authorino_controller.go#L723) function to assigned the value to the `roleBindingSearchName` variable

controllers/authorino_controller.go
```diff
@@ -443,73 +478,96 @@ func (r *AuthorinoReconciler) createAuthorinoPermission(authorino *api.Authorino
...
+			fmt.Errorf("failed to get %s binding for authorino ClusterRole, err: %v", roleBindingName, err),
 		)
-	}
+	} else {
+		// (cluster)rolebinding exists -> ensure it includes the serviceaccount among the subjects
 
-	// creates leader election role
-	// TODO: describe what this leader election is
-	return r.leaderElectionPermission(authorino, sa.GetName())
+		rb := authorinoResources.AppendSubjectToRoleBinding(roleBinding, serviceAccount) // TODO: Remove subject from the clusterrolebinding when the authorino instance is deleted. OwnerReference will not work for cluster-scoped resources :/
```
jjaferson, line 547: Nice one :)

controllers/authorino_controller.go
```diff
@@ -578,6 +636,39 @@ func (r *AuthorinoReconciler) createAuthorinoServices(authorino *api.Authorino)
 	return nil
 }
 
+func (r *AuthorinoReconciler) cleanupClusterScopedPermissions(ctx context.Context, crNamespacedName types.NamespacedName) {
+	crName := crNamespacedName.Name
+	sa := authorinoResources.GetAuthorinoServiceAccount(crNamespacedName.Namespace, authorinoServiceAccountName(crName))
```
jjaferson, line 641: If we move `authorinoServiceAccountName` to the resource package, we just need to pass the crName and namespace to the `GetAuthorinoServiceAccount`

jjaferson (commented):
I added a few comment regarding some small changes to the code other than that code looks good

controllers/authorino_controller.go
guicassolato: Some names are defined in  https://github.com/Kuadrant/authorino-operator/blob/main/pkg/resources/k8s_utils.go, while rbac, service, and deployment seem to be only for building basic structs based on k8s types.

Should I add it to rbac or to utils? Or maybe have a new file only for naming objects?

controllers/authorino_controller.go
guicassolato: Line 454 returns an error if the getting the service account fails to a reason other than a 404 (quite unusual). This one is for when the service account already exists (i.e., getting the service account did not fail), and therefore the process can move on.

controllers/authorino_controller.go
guicassolato: Sorry. I don't get it. This is to return the same errors as before. It's only that the logics to get/create the service account now moved to a separate function.

controllers/authorino_controller.go
jjaferson: True, what do you think if we add this function to utils, make it accessible only in the package then create a method in k8s_rbac.go that receives the name and namespace of the service account and return a service account's object with these attributes assigned?

controllers/authorino_controller.go
jjaferson: Not sure if I am missing something but I didn't understand the reason for adding the else statement, because if getting the services account doesn't fail then the method will return the retrieved service account anyway

guicassolato: What do you mean by "anyway"? The `else` statement _is_ where we return the SA when getting the SA does not fail.

controllers/authorino_controller.go
guicassolato: Done.

controllers/authorino_controller.go
guicassolato: It is not a `panic`, right? So we need to propagate the error.

controllers/authorino_controller.go
guicassolato: Done.

controllers/authorino_controller.go
guicassolato: And it's idempotent...and indifferent to `RoleBinding` or `ClusterRoleBinding` 😊

controllers/authorino_controller.go
guicassolato: Done.

jjaferson (approved):
Just tested the changes and the resources were created accordingly. Thanks for the PR @guicassolato

PR #23, opened 2022-01-13
Add OLM bundle

mikenairn:
Commit generated OLM bundle data and Dockerfile `make bundle`. 
Add CI jobs to execute verify-* targets.

Makefile updates:

* Add script to install a specific version of the operator SDK(v1.15.0) and use it in Makefile targets.
* Update bundle make target to output to default location (./bundle)
* Add make targets to verify actions that generate/modify code have been executed and output committed.

```
make help 
....
Verify
  verify-manifests      Verify manifests update.
  verify-bundle         Verify bundle update.
  verify-fmt            Verify fmt update.
```

### DIFF
--- a/bundle/manifests/authorino-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/authorino-operator.clusterserviceversion.yaml
@@ -1,0 +1,370 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  annotations:
+    alm-examples: |-
+      [
+        {
+          "apiVersion": "operator.authorino.kuadrant.io/v1beta1",
+          "kind": "Authorino",
+          "metadata": {
+            "name": "authorino-sample",
+            "namespace": "authorino-operator"
+          },
+          "spec": {
+            "clusterWide": true,
+            "image": "quay.io/3scale/authorino:latest",
+            "imagePullPolicy": "Always",
+            "listener": {
+              "port": null,
+              "tls": {
+                "certSecretRef": {
+                  "name": "authorino-cert"
+                },
+                "enabled": true
+              }
+            },
+            "oidcServer": {
+              "port": null,
+              "tls": {
+                "certSecretRef": {
+                  "name": "authorino-cert"
+                },
+                "enabled": true
+              }
+            },
+            "replicas": 1
+          }
+        }
+      ]
+    capabilities: Basic Install
+    categories: Integration & Delivery
+    containerImage: quay.io/3scale/authorino-operator:v0.0.1
+    createdAt: 2021-12-08T10-00-00Z
+    operators.operatorframework.io/builder: operator-sdk-v1.14.0+git
+    operators.operatorframework.io/project_layout: go.kubebuilder.io/v3
+    repository: https://github.com/Kuadrant/authorino-operator
+    support: kuadrant
+  name: authorino-operator.v0.0.1
+  namespace: placeholder
+spec:
+  apiservicedefinitions: {}
+  customresourcedefinitions:
+    owned:
+    - description: API to describe the desired protection for a service
+      displayName: AuthConfig
+      kind: AuthConfig
+      name: authconfigs.authorino.kuadrant.io
+      version: v1beta1
+    - description: API to create instances of authorino
+      displayName: Authorino
+      kind: Authorino
+      name: authorinos.operator.authorino.kuadrant.io
+      version: v1beta1
+  description: The operator to manage instances of Authorino
+  displayName: Authorino Operator
+  icon:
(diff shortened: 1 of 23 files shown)


bundle/manifests/authorino-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,370 @@
...
+                "enabled": true
+              }
+            },
+            "replicas": 1
+          }
+        }
+      ]
+    capabilities: Basic Install
+    categories: Integration & Delivery
+    containerImage: quay.io/3scale/authorino-operator:v0.0.1
```
mikenairn, line 42: I had changed this default image to quay.io/authorino/authorino-operator, but backed it out as it was likely to cause issues since there is no 0.0.1 tag on this repo and we seem to be setting the default version to 0.0.1 rather than latest here.

IMO we should be trying to move all the built images to the kuadrant quay org, but it's not what i was trying to do with this PR. If we do want to do that, we can do it separately.  

Do we have a script or doc describing what we do when we create a release of this operator? Also, what do we want the main branches of these operator repos pointing at? A released version or the latest version of the operator image?

guicassolato: > we seem to be setting the default version to 0.0.1 rather than latest here

I wish we were setting to latest, for just a brief moment TBH, but soon we will be issuing Authorino Operator v0.1.0 release and have the option to get best of both worlds: something more up-to-date with the current set of features stated in the manifests and yet more stable than latest.

(For something more general, please see my answer below about to which version main branches should be pointing to.)

> IMO we should be trying to move all the built images to the kuadrant quay org

+1

> Do we have a script or doc describing what we do when we create a release of this operator?

We do not. So far, release process has been very informal and lightweight, essentially driven to how we've been doing for https://github.com/kuadrant/authorino:
- First, CI ensures all merge commits to main branch are built and pushed to quay.io, with `latest` moved to the head of main branch
- Then, we issue a new release in GH pointing to a certain SHA-1 of the main branch. This usually coincides with the head of the branch and therefore with the `latest` image tag in quay.io. A new git tag is created with name "v0._X_._Y_" – major version 0 because we're not production-ready just yet, minor version _X_ to signal new RFEs potential breaking changes added, micro version _Y_ reserved for bug fixes.
- Finally, in quay.io, we manually create another "v0._X_._Y_" tag based on that same SHA-1 that was base for the release in GH and that is already available in quay.io.

Notice that the process doesn't say anything about the operator version stated in the Makefile and that end up in the bundle manifests. This is because the bundles are a relatively new thing in the Authorino Operator project, non-existent in the Authorino service whose release process is our base, and I guess we've just left the operator version to be manually bumped in a PR in preparation for a release, perhaps until we have a better process that can handle all those things more smoothly (i.e. operator version, operator image tag, and GH tag and release).

> what do we want the main branches of these operator repos pointing at? A released version or the latest version of the operator image?

IMO, it's always easier and more organised when main branches point to latest, but that means keeping branches for the releases instead of simple tags. The release process would have to change a bit.

Another option is using "develop" and "release" branches (aka Gitflow).

Operators are supposed be stable and changes in the API not to happen at an accelerated pace, but usually at chunks that you prepare for, and take time to sync with the operands... to announce to the community, etc. So either having branches for the releases or Gitflow should be manageable, in general.

On the other hand, changes do happen, and especially now that we are moving fast. Providing stable versions means teams will have to be releasing more often. The simpler this process is, the better for now.

Now trying to make sense on the options I see here (while still lacking some good automation in place)...

**No release branches / main pointing to a stable release:**
- No release process overhead
- Dev workflow sometimes might involve patching / changing manifests on-the-fly to pick latest unreleased changes

**No release branches / main branch pointing to `latest`:**
- No release process overhead
- Stable deployments will definitely involve patching / changing manifests on-the-fly to pick a release

**Release branches (forked from main as part of the release process):**
- main can default to `latest`
- release branches point to stable releases
- stable deployments and dev workflow simple and painless
- some release process overhead

**Gitflow:**
- main branch points to last stable release
- "develop" branch points to `latest`
- release branches are transient and created only for the release process (to set a fixed stable version and get incorporated to main)
- stable deployments and dev workflow simple and painless
- a lot more release process overhead

WDYT? Other options that I am not seeing.


cc @mikenairn @jjaferson @eguzki @maleck13 @rahulanand16nov

guicassolato (approved):

bundle/manifests/authorino-operator.clusterserviceversion.yaml
jjaferson: As we are still in development phase bugs and breaking changes are expected so I agree with @guicassolato in having the main branch pointing to latest. 

it can be helpful for testing as anyone using the `latest` image tag will be able to report bugs or unexpected behaviour from newly developed features, of course the latest image tag won't be stable but once we are confident enough with the changes that have been merged into the main branch we can then cut a release, tag the release and generate a new image referencing the release.

I am not really a fan of the Gitflow, I think it adds more complexity than it helps for instance we will need to keep track of at least two branches (dev and main), which in my point of view is not necessary.

Happy to hear what other thoughts are...

jjaferson (approved):

bundle/manifests/authorino-operator.clusterserviceversion.yaml
maleck13: Anyone willing to take on defining a release process that  identifies gaps we have , the components involved and the ideal e2e flow? I think we want to unify our process across each of the components

mikenairn: Agree that we should define and agree on a process that we can adopt for all repos we control, can also include how we expect the images to built and made available as well. I don't mind giving that a go.

jjaferson (approved):

guicassolato (approved):

bundle/manifests/authorino-operator.clusterserviceversion.yaml
eguzki: 3scale operators follow that `Release branches` and works for now. 

Operator's `main` branch should deploy operator's `latest` image.

PR #24, opened 2022-01-17
[WIP] Authorino manifest config

mikenairn:
**Authorino manifests as config**

Store authorino manifests in their own kustomization config file and include it in the default and install kustomizations. This allows an exact copy of the upstream version manifests to be stored giving us:
 * Easier comparisons (upstream vs local).
 * Ability to add kustomizations to those manifests if required.
 * Allows them to be included in different kustomizations
   (default vs install)
 * Tidier make targets that don't differ so much from the default, which
   coincidentally will alllow the "verify-bundle" make target to work
   again on CI.

Add new make targets for dealing with the different manifest types (authorino, install and deploy) and trigger them all as part of the top level "manifests" target.
Moved all manifest targets under the development section of the Makefile.
Removed install-authorino target since install target includes authorino manifests.

**Bundle generation and CI verification**

* Add missing role file to bundle.
* Add back "verify-bundle" CI job.

### DIFF
--- a/bundle/manifests/authorino-operator-authorino-authconfig-editor-role_rbac.authorization.k8s.io_v1_clusterrole.yaml
+++ b/bundle/manifests/authorino-operator-authorino-authconfig-editor-role_rbac.authorization.k8s.io_v1_clusterrole.yaml
@@ -2,7 +2,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   creationTimestamp: null
-  name: authorino-manager-role
+  name: authorino-operator-authorino-authconfig-editor-role
 rules:
 - apiGroups:
   - authorino.kuadrant.io
@@ -15,29 +15,9 @@ rules:
   - list
   - patch
   - update
-  - watch
 - apiGroups:
   - authorino.kuadrant.io
   resources:
   - authconfigs/status
   verbs:
   - get
-  - patch
-  - update
-- apiGroups:
-  - coordination.k8s.io
-  resources:
-  - leases
-  verbs:
-  - create
-  - get
-  - list
-  - update
-- apiGroups:
-  - ""
-  resources:
-  - secrets
-  verbs:
-  - get
-  - list
-  - watch
--- a/config/deploy/manifests.yaml
+++ b/config/deploy/manifests.yaml
@@ -1,3 +1,10 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    control-plane: controller-manager
+  name: authorino-operator
+---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
@@ -50,30 +57,20 @@ spec:
         description: AuthConfig is the schema for Authorino's AuthConfig API
         properties:
           apiVersion:
-            description: 'APIVersion defines the versioned schema of this representation
-              of an object. Servers should convert recognized schemas to the latest
-              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+            description: 'APIVersion defines the versioned schema of this representation of an object. Servers should convert recognized schemas to the latest internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
             type: string
           kind:
-            description: 'Kind is a string value representing the REST resource this
-              object represents. Servers may infer this from the endpoint the client
-              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+            description: 'Kind is a string value representing the REST resource this object represents. Servers may infer this from the endpoint the client submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
             type: string
           metadata:
             type: object
           spec:
-            description: Specifies the desired state of the AuthConfig resource, i.e.
-              the authencation/authorization scheme to be applied to protect the matching
-              service hosts.
+            description: Specifies the desired state of the AuthConfig resource, i.e. the authencation/authorization scheme to be applied to protect the matching service hosts.
             properties:
               authorization:
-                description: Authorization is the list of authorization policies.
-                  All policies in this list MUST evaluate to "true" for a request
-                  be successful in the authorization phase.
+                description: Authorization is the list of authorization policies. All policies in this list MUST evaluate to "true" for a request be successful in the authorization phase.
                 items:
-                  description: 'Authorization policy to be enforced. Apart from "name",
-                    one of the following parameters is required and only one of the
-                    following parameters is allowed: "opa", "json" or "kubernetes".'
+                  description: 'Authorization policy to be enforced. Apart from "name", one of the following parameters is required and only one of the following parameters is allowed: "opa", "json" or "kubernetes".'
                   oneOf:
                   - properties:
                       name: {}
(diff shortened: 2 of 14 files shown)


bundle/manifests/authorino-operator-authorino-authconfig-editor-role_rbac.authorization.k8s.io_v1_clusterrole.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   creationTimestamp: null
-  name: authorino-authconfig-editor-role
+  name: authorino-operator-authorino-authconfig-editor-role
```
mikenairn, line 5: The bundle includes the authorino cluster roles, but now that we are generating them via kustomize it's getting the namespace prefix we are setting in the default config. I don't think this is an issue, but if it we can still do it this way, but would need to use a different config other than default.

config/deploy/manifests.yaml
```diff
@@ -50,30 +57,20 @@ spec:
         description: AuthConfig is the schema for Authorino's AuthConfig API
         properties:
           apiVersion:
-            description: 'APIVersion defines the versioned schema of this representation
-              of an object. Servers should convert recognized schemas to the latest
-              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+            description: 'APIVersion defines the versioned schema of this representation of an object. Servers should convert recognized schemas to the latest internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
```
mikenairn, line 60: This seems to be down to the version of kustomize being used 3.5.4(https://github.com/Kuadrant/authorino/blob/main/Makefile#L73) vs 3.8.7 (https://github.com/Kuadrant/authorino-operator/blob/main/Makefile#L140)

mikenairn:
@guicassolato @jjaferson There was a missing file in the bundle i added which led me to look at this again and try and get the CI job working (make verify-bundle) which would have caught it ( wouldn't work via GH Actions with the current changes to the bundle make target for some reason). This could probably have been fixed in other ways, but i had started looking into how best to manage the operand(kuadrant-controller) from the kuadrant-operator and had been doing something similar to what I've done here to just keep the installation manifests in their own kustomize base that can be included by others. This seems, to me at least, a nicer way of doing things than joining multiple sets of manifests via commands in the Makefile.

Anyway, let me know what you think, no rush on this one.

bundle/manifests/authorino-operator-authorino-authconfig-editor-role_rbac.authorization.k8s.io_v1_clusterrole.yaml
guicassolato: This will require changes in the reconciler as it needs to know the name of these roles to create the bindings for the instance SAs: https://github.com/Kuadrant/authorino-operator/blob/26a1140e3e927cf567dbe1c54e98344e4ad681c6/controllers/authorino_controller.go#L51

guicassolato (changes requested):

bundle/manifests/authorino-operator-authorino-authconfig-editor-role_rbac.authorization.k8s.io_v1_clusterrole.yaml
mikenairn: OK, will look at doing this differently so it doesn't alter the names of the authorino resources.

guicassolato:
@mikenairn, IMO the manifests imported from https://raw.githubusercontent.com/Kuadrant/authorino/main/install/manifests.yaml should not change in the Authorino Operator repo via Makefile.

Authorino (service) owns the `AuthConfig` CRD and related RBAC, while Authorino Operator owns `Authorino` CRD and related RBAC. The reason why we decided to keep a copy of the Authorino manifests here was so one could install Authorino Operator and Authorino CRDs and RBAC and create a deployment of Authorino Operator, all in one single `kubectl apply` command. Other than that, those 2 sets of manifests could remain each one in its repo original repo only, where they are separately defined and maintained.

Say we [fix](https://github.com/Kuadrant/authorino-operator/pull/24#discussion_r786215998) the reconciler for the new names of the Authorino roles. The changes here proposed would still impact the dev workflow in the main Authorino repo for cases where the change there involves modifying the any of these resources. E.g., to add a new permission to the Authorino manager role, today you can change only in the Authorino repo; you don't even need to clone the Authorino Operator repo – just install/deploy the operator latest image and manifests, and in your branch for the Authorino repo, do the changes and re-apply on top.

mikenairn:
> @mikenairn, IMO the manifests imported from https://raw.githubusercontent.com/Kuadrant/authorino/main/install/manifests.yaml should not change in the Authorino Operator repo via Makefile.

Apart from the name prefix, which can be removed, they are unchanged. Only the means to which they are merged together for the OLM bundle and the deploy manifest file was updated.

> Authorino (service) owns the `AuthConfig` CRD and related RBAC, while Authorino Operator owns `Authorino` CRD and related RBAC. The reason why we decided to keep a copy of the Authorino manifests here was so one could install Authorino Operator and Authorino CRDs and RBAC and create a deployment of Authorino Operator, all in one single `kubectl apply` command. Other than that, those 2 sets of manifests could remain each one in its respective original repo only, where they are separately defined and maintained.
> 
> Say we [fix](https://github.com/Kuadrant/authorino-operator/pull/24#discussion_r786215998) the reconciler for the new names of the Authorino roles that kustomize is now prefixing. The changes here proposed would still impact the dev workflow in the main Authorino repo for cases where the change there involves modifying any of these resources. E.g., to add a new permission to the Authorino manager role, today you can change only in the Authorino repo; you don't even need to clone the Authorino Operator repo – just install/deploy the operator latest image and manifests, and in your branch for the Authorino repo, do the changes and re-apply on top.

Read this a few times, not totally sure i understand, but i think the problem is that i have added the authorino manifests to the install manifests of the operator? That's the only difference between this set of manifests and what was here previously, and can be undone. In my mind i just couldn't see the point of installing the operator CRDs without the authorino CRDs, but it's preferable we can leave it that way.

I can see as it is now that changes made in authorino would potentially require an update here due to us pulling the install manifests from the main branch, but that is true for what's on main now also, or am i just misunderstanding your point entirely?

guicassolato:
> Apart from the name prefix, which can be removed, they are unchanged. Only the means to which they are merged together for the OLM bundle and the deploy manifest file was updated.

If there is zero difference between the resources from the Authorino repo merged here to the Authorino Operator ones and how they are there in the Authorino repo originally, not even the names of the resources differ, then the means to which they are merged together do not matter and I am fine with the change.

> Read this a few times, not totally sure i understand, but i think the problem is that i have added the authorino manifests to the install manifests of the operator? That's the only functional difference between this set of manifests and what was here previously, and can be undone. In my mind i just couldn't see the point of installing the operator CRDs without the authorino CRDs, but if it's preferable we can leave it that way.

To me, having the Authorino manifests merged together into the Authorino Operator's ones under config/install feels a bit weird because those are resources that do not belong to the Authorino Operator repo, the types they refer to are not defined in this repo, and the operator does not technically require those resources to be applied to the cluster in order to be installed or deployed. That said, it is not a big deal though. In the end, you are right... To use Authorino, users do need to install both sets of manifests. It's more about the what it means (at least in my mind):
- **Install the operator:** apply the `Authorino` CRD plus Authorino Operator RBAC
- **Deploy the operator:** create an Authorino Operator `Deployment`
- **Install Authorino:** apply the `AuthConfig` CRD plus Authorino role definitions (just cluster roles, no bindings!)
- **Deploy Authorino:** create an `Authorino` CR, which causes a running instance of the operator to create the Authorino `Deployment`, `Service`, `ServiceAccount`, `RoleBinding` and `ClusterRoleBinding`s
 
> I can see as it is now that changes made in authorino would potentially require an update here due to us pulling the install manifests from the main branch, but that is true for what's on main now also, or am i just misunderstanding your point entirely?

I guess you got it right. Whenever the Authorino manifests change in the Authorino repo, we need to send a PR to the operator, to update the handy `config/deploy/manifests.yaml` file that we host here (for simple one-line installation and deployment of the Authorino Operator's and Authorino's CRDs)... Since #23 was merged, to update the bundle manifests as well. (The contrary is not true obviously – i.e. changes in the operator manifests do not require any PR to the main Authorino repo.)

An idea we discussed when deciding about this organisation for the manifests was of bringing Authorino's `AuthConfig` and RBAC definitions all to the operator code base, and remove them from Authorino's one. This repo would define all the types and generate all the manifests. Authorino (the service) would still need the types though, so it would to have to import the Go packages, but no k8s resource YAML file hosted there. In the end, we went with the current separation to respect each component's responsibilities and relative independence one from another, that kind of exist to some extent. We can revisit this decision.

PR #25, opened 2022-01-17
Update Authorino manifests

guicassolato:


utils/install-operator-sdk.sh
```diff
@@ -5,7 +5,7 @@
 
 set -euo pipefail
 
-ARCH=$(case $(uname -m) in x86_64) echo -n amd64 ;; aarch64) echo -n arm64 ;; *) echo -n "$(uname -m)" ;; esac)
+case $(uname -m) in x86_64|aarch64) ARCH="amd64";; *) ARCH="$(uname -m)";; esac
```
mikenairn, line 8: fyi, I'm going to add a CI test to make sure these work on both macOS and Linux. wip here https://github.com/Kuadrant/kuadrant-operator/runs/4846610978?check_suite_focus=true

Couple of other issues with the script i noticed, will fix it up and update it across the repos it's been added to.

mikenairn:
@guicassolato I wondered how i had managed to commit a bundle that was missing files, but it's obviously because we are tracking the main branch of authoino, and that's been updated since i added it. Does it make sense to track main, or should we fixing the operator to a specific version of authorino? Couple of problems with tracking main that i can see:
* verify-manifests and verify-bundle CI jobs are going to break on PRs of this repo when someone updates something in the authorino that affects the manifests pulled into here.
* We are likely going to break the operator installation at some point just by updating authorino

In the past we would have always fixed a version of the operand in the operator meaning a version of the operator always works with a specific version of it's operand. Would that not make more sense here? It doesn't stop you from overriding it in development, just the default isn't something that is going to change under your feet.

mikenairn (approved):

guicassolato:
> @guicassolato I wondered how i had managed to commit a bundle that was missing files, but it's obviously because we are tracking the main branch of authoino, and that's been updated since i added it. Does it make sense to track main, or should we fixing the operator to a specific version of authorino? Couple of problems with tracking main that i can see:
> 
>     * verify-manifests and verify-bundle CI jobs are going to break on PRs of this repo when someone updates something in authorino that affects the manifests pulled into here.
> 
>     * We are likely going to break the operator installation at some point just by updating authorino
> 
> 
> In the past we would have always fixed a version of the operand in the operator meaning a version of the operator always works with a specific version of it's operand. Would that not make more sense here? It doesn't stop you from overriding it in development, just the default isn't something that is going to change under your feet.

Since we seem to have [decided](https://github.com/Kuadrant/authorino-operator/pull/23#discussion_r784712447) for keeping release branches to fix a version to the operator (while still defaulting to `latest` in the main branch), perhaps we could also fix the version of Authorino in the release branches and keep fetching from Auhtorino `main` in the operator main branch?

In the case of PRs that you mentioned, maybe it's a good thing, kind of a warning that operator main is outdated because of a change in Authorino main and that we should update the manifests ASAP (in a separate PR) and rebase all work pending to be merged? WDYT?

mikenairn:
> Since we seem to have [decided](https://github.com/Kuadrant/authorino-operator/pull/23#discussion_r784712447) for keeping release branches to fix a version to the operator (while still defaulting to `latest` in the main branch), perhaps we could also fix the version of Authorino in the release branches and keep fetching from Auhtorino `main` in the operator main branch?
> 
> In the case of PRs that you mentioned, maybe it's a good thing, kind of a warning that operator main is outdated because of a change in Authorino main and that we should update the manifests ASAP (in a separate PR) and rebase all work pending to be merged? WDYT?

I guess it comes down how you look at the relationship between the operator(authorino-operator) and it's operand (authorino). To me an operator has a dependency on an operand, and is written to work with a particular version of it, and the development of either is completely separate and can be worked on without affecting the other. 

Having the main branch here default to a latest build of the operator which itself is built from the main branch (the product of this repo), makes sense to me, but i wouldn't really think of the operand which is coming from a completely different repo, with it's own development processes and releases, in the same way.

That said, my thinking here could be skewed by the fact i have never worked on an operator where we actually control the operand as well and maybe there is some advantage to having them more tightly coupled.

PR #26, opened 2022-01-18
Default operator image to latest

guicassolato:
To generate deploy manifests and bundle for the main branch:

```sh
make manifests bundle
```

To generate deploy manifests and bundle for a release branch:

```sh
make manifests bundle OPERATOR_VERSION=0.1.0 AUTHORINO_VERSION=0.6.0
```


jjaferson (approved):
One thing I noticed is that when generate a new bundle the [containerImage](https://github.com/Kuadrant/authorino-operator/blob/main/bundle/manifests/authorino-operator.clusterserviceversion.yaml#L42) metadata is not changing to reflect the operatorImage.

PR #27, opened 2022-01-18
Update operator-sdk install script

mikenairn:
Update operator-sdk install script to work on both linux and mac.
* Create target directory if it doesn't exist.
* Use `shasum` to check checksums on macOS.

Add CI job to check scripts required in local development environments can execute on both linux and macOS.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -37,3 +37,25 @@ jobs:
       - name: Run make verify-fmt
         run: |
           make verify-fmt
+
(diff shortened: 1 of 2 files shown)


mikenairn:
@guicassolato I believe this should work on a mac fine, the CI test is passing at least, but would you mind confirming before i update this in other repos?

guicassolato (approved):
Tested in MacOS x86_64. It seems to work perfectly.
Thanks @mikenairn!

Issue #28, opened 2022-01-20
Does not reconcile log level changes


eguzki:
### Steps to reproduce
* Create Authorino CR with no log level spec
* Update CR adding `spec.logLevel = 'debug'`

Authorino is not rolled out with the new conf.

jjaferson:
`logLevel` and `logMode` are added as env variables to the deployment resource of Authorino and if they are not added when the CR is created, the operator is not detecting the change in the deployment as the env vars don't exist in the deployment, therefore the update will not work. 

A quick fix to this is to patch the authorino CR to add the desired env variable. 
```bash
kubectl set env deployment/<authorino-cr-name> LOG_LEVEL=debug  -n <authorino-cr-namespace>
```

I will create a PR to fix this issue and link here.

PR #30, opened 2022-01-21
Fixes an issue where the operator does not reconcile env variables

jjaferson:
This PR fixes an issue where the operator was not updating env variables to the deployment resources.

Fixes: https://github.com/Kuadrant/authorino-operator/issues/28


guicassolato (approved):
LGTM

PR #29, opened 2022-01-20
[WIP] Adds rbac proxy sidecar container to authorinos instance

jjaferson:
Adds rbac sidecar proxy to protect authorinos metric endpoint

Closes: https://github.com/Kuadrant/authorino-operator/issues/4

PR #34, opened 2022-02-28
Changes authorino service metric port

jjaferson:
As result of [removing proxy sidecar container](https://github.com/Kuadrant/authorino-operator/pull/29) from the authorino instances, there authorino metric service needs to point to the default http port exposed by authorino.

* [Default port exposed by authorino](https://github.com/Kuadrant/authorino/blob/main/main.go#L75 )

### DIFF
--- a/pkg/resources/k8s_services.go
+++ b/pkg/resources/k8s_services.go
@@ -2,7 +2,6 @@ package resources
 
 import (
 	k8score "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
 // GetAuthorinoServices returns all the service required by an instance of authorino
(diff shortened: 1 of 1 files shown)


pkg/resources/k8s_services.go
```diff
@@ -36,9 +36,9 @@ func newAuthService(authorinoName, serviceNamespace string) *k8score.Service {
 func newMetricService(authorinoName, serviceNamespace string) *k8score.Service {
 	serviceName := "controller-metrics"
 	return newService(serviceName, serviceNamespace, authorinoName, k8score.ServicePort{
-		Name:       "https",
-		Port:       8443,
-		TargetPort: intstr.FromString("https"),
+		Name:       "http",
+		Port:       8080,
+		TargetPort: intstr.FromString("http"),
```
guicassolato, line 41: Does the pod exports a port named "http"?

jjaferson: Not really, there are no named ports exposed in the pod. I think we can remove this TargetPort and if needed in the future expose the metrics port via named port in the pod, wdyt?

guicassolato (approved):
❤️

Issue #42, opened 2022-04-22
Better name for the operator workload


guicassolato:
Authorino Operator's pod is named `authorino-operator-controller-manager`, as bootstrapped by default in projected created using the [Operator SDK](https://sdk.operatorframework.io).

https://github.com/Kuadrant/authorino-operator/blob/785b0777ac031fbd2a10cfa5478831b8db9419b2/config/manager/manager.yaml#L11
https://github.com/Kuadrant/authorino-operator/blob/785b0777ac031fbd2a10cfa5478831b8db9419b2/config/default/kustomization.yaml#L9

This is a long name that could be better simplified to just `authorino-operator`.

PR #52, opened 2022-05-04
Renames resources to authorino-operator

jjaferson:
Closes: https://github.com/Kuadrant/authorino-operator/issues/42


controllers/authorino_controller.go
```diff
@@ -67,6 +67,8 @@ const (
 // +kubebuilder:rbac:groups="*",resources=clusterrolebindings,verbs=get;list;watch;create;update;
 // +kubebuilder:rbac:groups="*",resources=serviceaccounts,verbs=get;list;watch;create;update;
 // +kubebuilder:rbac:groups="*",resources=roles,verbs=get;list;watch;create;update;
+// +kubebuilder:rbac:groups="*",resources=tokenreviews,verbs=create;
+// +kubebuilder:rbac:groups="*",resources=subjectaccessreviews,verbs=create;
```
guicassolato, line 71: Why including these in the Authorino controller? It looks like these are permissions only needed for kube-rbac-proxy, whose roles, binding, and service definitions (!) would be removed with this PR.

Nothing against disabling kube-rbac-proxy that is now protecting the metrics endpoint, but perhaps in this case we want to also remove the patch on the `authorino-operator` deployment. Maybe do this part of the change in a separate PR?

As-is, I think we're removing bits of the metrics service, while leaving other stuff behind, e.g., the sidecar, the permissions for the process to issue tokenreviews and subjectaccessreviews, no?

guicassolato: These are not kube-rbac-proxy related, @jjaferson. Authorino too needs to create tokenreviews and subjectaccessreviews, to implement the features of Kubernetes authn and Kubernetes authz.

jjaferson: This is because authorino operator needs these permission in order to be able to grant them to the authorino instances.

See below the erro message returned from the operator

```
 failed to create authorino-k8s-auth binding for authorino ClusterRole, err: clusterrolebindings.rbac.authorization.k8s.io "authorino-k8s-auth" is forbidden: user "system:serviceaccount:authorino-operator:authorino-operator" (groups=["system:serviceaccounts" "system:serviceaccounts:authorino-operator" "system:authenticated"]) is attempting to grant RBAC permissions not currently held:
      {APIGroups:["authentication.k8s.io"], Resources:["tokenreviews"], Verbs:["create"]}
      {APIGroups:["authorization.k8s.io"], Resources:["subjectaccessreviews"], Verbs:["create"]}
    reason: UnableToBindingForClusterRole
```

guicassolato (dismissed):

guicassolato (dismissed):

guicassolato (approved):

PR #62, opened 2022-05-23
Authorino env vars and command-line flags

guicassolato:
- Refactor names of the constants that hold env var names, for consistency
- Add `spec.deepMetricsEnabled: bool` -> `DEEP_METRICS_ENABLED`
- Add `spec.listener.timeout: int` -> `TIMEOUT`
- General refactoring of the usage the env var refs (impl and tests), for consistency
- Add `spec.metrics.port` ~> `--metrics-addr=:port`
- `spec.deepMetricsEnabled` -> `spec.metrics.deep`
- Set `--enable-leader-election` when `spec.replicas > 1`

### Verification steps

```sh
kind create cluster --name authorino
make install deploy OPERATOR_IMAGE=authorino-operator:local
```

```sh
# defaults
kubectl apply -f -<<EOF
kind: Authorino
metadata:
  name: authorino
spec:
  listener:
    tls:
      enabled: false
  oidcServer:
    tls:
      enabled: false
EOF
```

```sh
# metrics
kubectl apply -f -<<EOF
kind: Authorino
metadata:
  name: authorino
spec:
  listener:
    tls:
      enabled: false
  oidcServer:
    tls:
      enabled: false
  metrics:
    port: 5000
    deep: true
EOF
```

```sh
# leader election and authz timeout
kubectl apply -f -<<EOF
kind: Authorino
metadata:
  name: authorino
spec:
  replicas: 2
  listener:
    tls:
      enabled: false
    timeout: 500
  oidcServer:
    tls:
      enabled: false
  metrics:
    port: 5000
    deep: true
EOF
```


eguzki (approved):

eguzki:
Question, for curiosity. I get that multiple replicas can be used to scale horizontally for the gRPC (and recent HTTP auth endpoints). But how does the leader election work? I mean, N replicas are getting events on changes in authConfigs? IF so, what is the meaning of leader election. i.e., what can the leader do that others cannot do?

eguzki:
Thanks for the explanation, fair enough

PR #71, opened 2022-06-20
README enhancements

### LINKED ISSUE #67: Latest changes in the Authorino API not reflected in the README
A few latest changes in the Authorino CR have not been reflected in the README. This is the case of the new fields such as `spec.metrics`, `spec.listener.ports` and `spec.listener.timeout`.

guicassolato:
Closes #67.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,128 +1,72 @@
 # Authorino Operator
 
+A Kubernetes Operator to manage [Authorino](https://github.com/Kuadrant/authorino) instances.
+
 [![License](https://img.shields.io/badge/license-Apache--2.0-blue.svg)](http://www.apache.org/licenses/LICENSE-2.0)
-
-## Overview
-
-A Kubernetes Operator to manage [Authorino](https://github.com/Kuadrant/authorino) deployments.
-
-## Custom Resources
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,162 +1,251 @@
...
+    <tr>
+      <th>Catalog</th>
+      <td><a href="https://quay.io/kuadrant/authorino-operator-catalog">quay.io/kuadrant/authorino-operator-catalog</a></td>
+    </tr>
+  </tbody>
+</table>
 
-Example:
-```yaml
+1. Create the namespace for the Operator
```
eguzki, line 61: In bare kubernetes, OLM is not installed. Maybe add a section to install the OLM? `operator-sdk` CLI has a command for that:

```
operator-sdk olm install
```

eguzki (approved):

README.md
guicassolato: There's a [Quick Start](https://olm.operatorframework.io/docs/getting-started/) section in the OLM page we linked above. It covers installing OLM in a cluster, general instructions to install operators using OLM, as well as the prerequisite of having the SDK binary itself somewhere locally (another step that might not be obvious for bare Kubernetes users, IMO).

So, given that there are not one but actually 2 steps to get OLM in the K8s cluster, the fact that in some cases OLM will be already available in the cluster (e.g. OpenShift users, K8s users already familiar with Operator SDK and OLM), plus the existence of a good page dedicated to such part of the instructions, I think I prefer saving this step and keep our sequence simple.

Thanks for the good tip anyway, @eguzki! I'll leave it here together with this reply of mine for the record, in case somebody else is wondering.

PR #84, opened 2022-07-21
Fix tests

jjaferson:
Since the PR for [preflights checks](https://github.com/Kuadrant/authorino-operator/pull/22) was added as requirement for when TLS is enabled the unit tests were failing due to a mock of the TLS cert secret not being created. 

This PR addresses:

- [x] The creation of a mock TLS cert secret for unit tests
- [x] Fix an issue where errors were being returned when the pre-flight check fails

### DIFF
--- a/controllers/authorino_controller_test.go
+++ b/controllers/authorino_controller_test.go
@@ -48,7 +48,7 @@ var _ = Describe("Authorino controller", func() {
 				err := k8sClient.Get(context.TODO(),
 					nsdName,
 					&authorino)
-				return err == nil && isAuthorinoInstanceReady(&authorino)
+				return err == nil && authorinoInstance.Status.Ready()
 			}, timeout, interval).Should(BeFalse())
 		})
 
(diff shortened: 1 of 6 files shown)


controllers/authorino_controller_test.go
```diff
@@ -299,3 +325,13 @@ func checkAuthorinoEnvVar(authorinoInstance *api.Authorino, envs []k8score.EnvVa
 		}
 	}
 }
+
+func isAuthorinoInstanceReady(authorino *api.Authorino) bool {
```
guicassolato, line 329: In Authorino we went with defining a function of the Status type for this:
https://github.com/Kuadrant/authorino/blob/0a5c235f512d4be08638665518a3c4a18a9e9532/api/v1beta1/auth_config_types.go#L668

jjaferson: Good idea

guicassolato (approved):
Excellent! Thanks @jjaferson

PR #86, opened 2022-08-02
Option to enable/disable the prevention of host collisions

guicassolato:
Following https://github.com/Kuadrant/authorino/pull/340.


guicassolato:
The PR in https://github.com/kuadrant/authorino that motivated this change was closed.

PR #88, opened 2022-08-23
update security context to drop all capabilities & set seccomp 

akoserwal:
Fix:
Warning: would violate PodSecurity "restricted:v1.24": unrestricted capabilities (container "manager" must set securityContext.capabilities.drop=["ALL"]), seccompProfile (pod or container "manager" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")

* `container "manager" must set securityContext.capabilities.drop=["ALL"]`: Drop all capabilities and add only those required 

* https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-seccomp-profile-for-a-container

### DIFF
--- a/bundle/manifests/authorino-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/authorino-operator.clusterserviceversion.yaml
@@ -283,6 +283,12 @@ spec:
                 resources: {}
                 securityContext:
                   allowPrivilegeEscalation: false
+                  capabilities:
+                    drop:
+                    - ALL
+                  readOnlyRootFilesystem: true
+                  seccompProfile:
+                    type: RuntimeDefault
               securityContext:
                 runAsNonRoot: true
               serviceAccountName: authorino-operator
--- a/config/deploy/manifests.yaml
+++ b/config/deploy/manifests.yaml
@@ -2568,6 +2568,12 @@ spec:
           periodSeconds: 10
         securityContext:
           allowPrivilegeEscalation: false
+          capabilities:
+            drop:
+            - ALL
+          readOnlyRootFilesystem: true
+          seccompProfile:
+            type: RuntimeDefault
       securityContext:
         runAsNonRoot: true
       serviceAccountName: authorino-operator
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -33,6 +33,12 @@ spec:
         name: manager
         securityContext:
           allowPrivilegeEscalation: false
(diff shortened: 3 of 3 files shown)


guicassolato:
Hi, @akoserwal. Thanks for this PR!

Do you mind running `make bundle` to also update the bundle manifests, please?

bundle/manifests/authorino-operator.clusterserviceversion.yaml
```diff
@@ -283,6 +283,12 @@ spec:
                 resources: {}
                 securityContext:
                   allowPrivilegeEscalation: false
+                  capabilities:
+                    drop:
+                    - ALL
```
guicassolato, line 288: Just wondering... We already run as non-root as escalation of privilege is not allowed. Isn't `capabilities.drop: ["ALL"]` redundant in this case?

```sh
$ docker run --cap-add SYS_ADMIN --user 1001 --entrypoint /bin/grep quay.io/kuadrant/authorino-operator:latest -- Cap /proc/self/status
CapInh:	0000000000000000
CapPrm:	0000000000000000
CapEff:	0000000000000000
CapBnd:	00000000a82425fb
CapAmb:	0000000000000000

$ docker run --cap-add SYS_ADMIN --user 1001 --entrypoint /bin/grep <image-based-on-this-branch> -- Cap /proc/self/status
CapInh:	0000000000000000
CapPrm:	0000000000000000
CapEff:	0000000000000000
CapBnd:	00000000a82425fb
CapAmb:	0000000000000000
```

akoserwal: Yes, I agree. when I was deploying the operator, I got this warning

```
Warning: would violate PodSecurity "restricted:v1.24": unrestricted capabilities (container "manager" must set securityContext.capabilities.drop=["ALL"]), seccompProfile (pod or container "manager" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost") 
```
I think "dropping all capabilities" will only reduce any further risk of remote code execution, which seems more like a good-to-have thing. run as non-root already handle it. When I added both recommendations, it suppressed the warning.

guicassolato: > When I added both recommendations, it suppressed the warning.

Good enough for me.

Thanks, @akoserwal!

guicassolato (approved):

guicassolato (approved):

PR #90, opened 2022-08-25
Embed operator version and authorino image version

didierofrivia:
TODO


guicassolato:
Closing this for now until we can revisit it with a solution to get the Authorino version to default the operand image to injected into the binary, when building the operator image in CI (automated workflows). A possible approach can be fetching from the Authorino repo the latest available release.

cc @didierofrivia

Issue #82, opened 2022-07-14
Allow for adding labels to service/deployments created by operator


OperationalDev:
It would be nice if there was a way to add labels to objects created by the operator when we create an authorino object.

This could be as simple as propagating the the labels on the authorino object to all of the objects created by the operator for this object.

PR #91, opened 2022-08-25
Propagating Authorino labels

didierofrivia:
This PR makes it possible to propagate the labels defined in the Custom Resource `Authorino` to all the Kubernetes objects the operator creates.

Closes https://github.com/Kuadrant/authorino-operator/issues/82

## Verification Steps

1. Create cluster
```bash
kind create cluster
```

2. Create namespace for the operator
```bash
```

3. Install manifests 
```bash
make install
```

4. Run the operator
```bash
make run
```

5. In another terminal apply the following `Authorino` CR
```bash
kubectl apply  -f -<<EOF
kind: Authorino
metadata:
  name: authorino
  labels:
    lola: mento
spec:
  image: quay.io/kuadrant/authorino:latest
  replicas: 1
  clusterWide: true
  listener:
    tls:
      enabled: false
  oidcServer:
    tls:
      enabled: false
EOF
```

6. Find the Kubernetes objects created by filtering that level
```bash
kubectl get services -l 'lola in (mento)'

NAME                                TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)              AGE
authorino-authorino-authorization   ClusterIP   10.96.134.186   <none>        50051/TCP,5001/TCP   5m2s
authorino-authorino-oidc            ClusterIP   10.96.142.39    <none>        8083/TCP             5m2s
authorino-controller-metrics        ClusterIP   10.96.140.39    <none>        8080/TCP             5m2s

```

```bash
kubectl get deployments -l 'lola in (mento)'

NAME        READY   UP-TO-DATE   AVAILABLE   AGE
authorino   1/1     1            1           6m43s
```

7. PROFIT! Now you can delete the cluster
```bash
kind delete cluster
```

### DIFF
--- a/controllers/authorino_controller.go
+++ b/controllers/authorino_controller.go
@@ -177,7 +177,7 @@ func (r *AuthorinoReconciler) getAuthorinoInstance(namespacedName types.Namespac
 	if err != nil {
 		if errors.IsNotFound(err) {
 			r.Log.Info("Authorino CR not found.")
-			r.cleanupClusterScopedPermissions(context.Background(), namespacedName)
+			r.cleanupClusterScopedPermissions(context.Background(), namespacedName, authorinoInstance.Labels)
 			return nil, nil
 		}
 		return nil, err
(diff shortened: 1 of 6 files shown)


guicassolato (approved):
This looks good.

Important to be aware that there is NO guarantee that other resources, created by other controllers based on these first-level resources created by the Authorino Operator, will have the same labels as the Authorino CR. This is the case, for example, of pods created from the Authorino Deployment; the pods will not share the same labels as Deployment. This is fine because, as long as the label selectors in the Service and in the Deployment, as well as label template for the pods stated in the Deployment continue to match – and those are not affected by this PR! – there's no risk of label mismatch in the Kubernetes pod autoscaling controller and service networking.

PR #95, opened 2022-08-30
Minimal example of the Authorino CR in the ClusterServiceVersion

guicassolato:
So we show no more than what's strictly needed. ~The only exception is `tls.enabled: true`.~

```yaml
kind: Authorino
metadata:
  name: authorino-sample
spec:
  listener:
    tls:
      enabled: false
  oidcServer:
    tls:
      enabled: false
```

### DIFF
--- a/bundle/manifests/authorino-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/authorino-operator.clusterserviceversion.yaml
@@ -13,18 +13,12 @@ metadata:
           "spec": {
             "listener": {
               "tls": {
-                "certSecretRef": {
-                  "name": "authorino-server-cert"
-                },
-                "enabled": true
+                "enabled": false
               }
             },
             "oidcServer": {
               "tls": {
-                "certSecretRef": {
-                  "name": "authorino-oidc-server-cert"
-                },
-                "enabled": true
+                "enabled": false
               }
             }
           }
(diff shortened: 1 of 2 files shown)


bundle/manifests/authorino-operator.clusterserviceversion.yaml
```diff
@@ -8,32 +8,25 @@ metadata:
...
-            "image": "quay.io/kuadrant/authorino:latest",
-            "imagePullPolicy": "Always",
             "listener": {
-              "port": null,
               "tls": {
                 "certSecretRef": {
-                  "name": "authorino-cert"
+                  "name": "authorino-server-cert"
                 },
                 "enabled": true
```
guicassolato, line 19: Do we want `enabled: false` here, make it even smaller without `certSecretRef` and nothing else required to try the example, or would that be sending a wrong message that we recommend TLS to be disabled, which we obviously do not?

In the opposite direction, `enabled: true` could be omitted here. But could that impose an obstacle to users who want to try without TLS first (e.g. for non-prod env) but think the only way is by creating the TLS secrets?

didierofrivia: I'd say `"enabled": false"`. Thus the config not only is smaller but also communicates to the user the setting exists.

guicassolato: Done

didierofrivia (approved):

PR #111, opened 2023-04-20
Tighten API group specification for RBAC

willthames:
When possible it's much better to explicitly specify RBAC constraints, in particular for APIGroups, Resources and Verbs where the set of possibilities is limited to start with

Specify the API group for all RBAC permission requirements.

### DIFF
--- a/bundle/manifests/authorino-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/authorino-operator.clusterserviceversion.yaml
@@ -58,27 +58,7 @@ spec:
       clusterPermissions:
       - rules:
         - apiGroups:
-          - '*'
-          resources:
-          - clusterrolebindings
-          verbs:
-          - create
-          - get
-          - list
-          - update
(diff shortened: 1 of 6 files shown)


willthames:
There is an outstanding question of whether events should be in the core API group (i.e. `""`) or the new events type `events.k8s.io`. For most event consumption purposes, the core group is likely correct but if you want to publish events `events.k8s.io` seems to be more common.

guicassolato:
Hi @willthames. Your PR looks good and I'll be glad to merge it. Thank you so much for your contribution!

Would you mind running `make bundle`? It's just so it updates the CSV too, at `bundle/manifests/authorino-operator.clusterserviceversion.yaml`.

Cheers!

willthames:
Thanks @guicassolato - I've run `make bundle` now and re-pushed.

guicassolato (approved):
Thank you so much for this @willthames!

PR #113, opened 2023-05-10
SCC supplemental group

guicassolato:
Related to https://github.com/Kuadrant/authorino/pull/391


guicassolato:
I guess we won't be needing this after moving https://github.com/Kuadrant/authorino/pull/391 to **OPTION 4**, i.e. Authorino home directory owned by `authorino:root` with `750` permissions applied.

PR #122, opened 2023-07-18
readme: format tables & deploy authorino instance to authorino-operator namespace in example

KevFan:
Minor `README` improvements:

* Format tables so that spacing is consistent (raw markdown will be easier to read for the tables)
* Deploy authorino instance to `authorino-operator` instead of `myapi`
  * `myapi` namespace is not explictly reference / created in the `README`

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,10 +38,10 @@ make deploy
   ```
 </details>
 
-
 ### Installing via OLM
 
-To install the Operator using the [Operator Lifecycle Manager](https://olm.operatorframework.io/), you need to make the Operator CSVs available in the cluster by creating a `CatalogSource` resource.
+To install the Operator using the [Operator Lifecycle Manager](https://olm.operatorframework.io/), you need to make the
+Operator CSVs available in the cluster by creating a `CatalogSource` resource.
 
 The bundle and catalog images of the Operator are available in Quay.io:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -85,7 +86,7 @@ EOF
 Once the Operator is up and running, you can request instances of Authorino by creating `Authorino` CRs. E.g.:
 
 ```sh
-kubectl -n myapi apply -f -<<EOF
+kubectl -n authorino-operator apply -f -<<EOF
```
guicassolato, line 89: This will not work. The idea here is to highlight that Authorino can be deployed as well as a dedicated instance within each application namespace.

Authorino deployed to the `authorino-operator` ns makes sense when using Authorino in cluster-wide mode, which is not the case of the exemplified CR.

KevFan: hmm, that's fair. I'm happy to revert it back to `myapi` if that's the preference but I would prefer if the namespace exists 🤔

Any thoughts on using the `default` namespace ?

guicassolato: `default` is fine. In fact, it would match most of the tutorials.

KevFan: Cool, I've pushed the change to use `default` namespace instaad 👍

README.md
```diff
@@ -38,10 +38,10 @@ make deploy
   ```
 </details>
 
-
 ### Installing via OLM
 
-To install the Operator using the [Operator Lifecycle Manager](https://olm.operatorframework.io/), you need to make the Operator CSVs available in the cluster by creating a `CatalogSource` resource.
+To install the Operator using the [Operator Lifecycle Manager](https://olm.operatorframework.io/), you need to make the
+Operator CSVs available in the cluster by creating a `CatalogSource` resource.
```
guicassolato, line 44: I guess your IDE doesn't automatically word wrap markdown content while keeping it in a single line of code.

guicassolato (approved):

README.md
KevFan: Yes. It looks to add a hard break at 120 characters also when auto reformatting on Goland IDE 🤔 

Happy to revert these anyway if that's the preference. It's pretty minor

PR #123, opened 2023-07-19
Add release workflow

### LINKED ISSUE #114: Set the `replaces` directive in the CSV
Release branches miss the `replaces` directive in the ClusterServiceVersion (CSV) manifest, set manually in the pull requests to the operator hubs, e.g.
- https://github.com/redhat-openshift-ecosystem/community-operators-prod/blob/19da6a06ab1db0351c0f354827915957830c59f4/operators/authorino-operator/0.7.0/manifests/authorino-operator.clusterserviceversion.yaml#L350
- https://github.com/k8s-operatorhub/community-operators/blob/9e7f3d392a50b17426a5c08ddb5b15f26dc2193b/operators/authorino-operator/0.7.0/manifests/authorino-operator.clusterserviceversion.yaml#L350

The directive should be filled whenever there's a new release of the operator.

grzpiotrowski:
- Added a new make target fix-csv-replaces
- Added a new make target prepare-release
- Moved tool installation to a different section in the Makefile
- Added the Release operator workflow

Fixes #114

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,53 @@
+name: Release operator
+
+on:
+  workflow_dispatch:
+    inputs:
+      commitSha:
+        description: Commit SHA
+        required: true
+      operatorVersion:
+        description: Operator version
+        required: true
+      authorinoVersion:
+        description: Authorino version
+        required: true
+        default: latest
+
+jobs:
+  build:
+    name: Release operator
+    runs-on: ubuntu-20.04
+    steps:
+    - name: Install gettext-base
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y gettext-base
+    - name: Set up Go 1.19.x
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.19.x
+      id: go
+    - name: Checkout code at commit SHA
+      uses: actions/checkout@v3
+      with:
+        ref: ${{ github.event.inputs.commitSha }}
+    - name: Create release branch
+      run: |
+        git checkout -b release-v${{ github.event.inputs.operatorVersion }}
+    - name: Prepare release
+      run: VERSION=${{ github.event.inputs.operatorVersion }} AUTHORINO_VERSION=${{ github.event.inputs.authorinoVersion }} GH_TOKEN=${{ secrets.GITHUB_TOKEN }} make prepare-release
+    - name: Commit and push
+      run: |
+        git config --global user.name "github-actions[bot]"
+        git config --global user.email "github-actions[bot]@users.noreply.github.com"
+        git commit -m "Prepared release v${{ github.event.inputs.operatorVersion }}" -a
+        git push origin release-v${{ github.event.inputs.operatorVersion }}
+    - name: Create release
+      uses: softprops/action-gh-release@v1
+      with:
+        name: v${{ github.event.inputs.operatorVersion }}
+        tag_name: v${{ github.event.inputs.operatorVersion }}
+        body: "**This release enables installations of Authorino v${{ github.event.inputs.authorinoVersion }}**"
+        generate_release_notes: true
+        target_commitish: release-v${{ github.event.inputs.operatorVersion }}
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -3,8 +3,8 @@ name: Release operator
 on:
   workflow_dispatch:
     inputs:
-      commitSha:
-        description: Commit SHA
+      gitRef:
+        description: Commit SHA or branch name
         required: true
       operatorVersion:
         description: Operator version
@@ -13,6 +13,10 @@ on:
         description: Authorino version
         required: true
         default: latest
+      prerelease:
+        description: Is the release a prerelease
+        required: false
+        type: boolean
 
 jobs:
   build:
@@ -28,11 +32,12 @@ jobs:
       with:
         go-version: 1.19.x
       id: go
-    - name: Checkout code at commit SHA
+    - name: Checkout code at git ref
       uses: actions/checkout@v3
       with:
-        ref: ${{ github.event.inputs.commitSha }}
+        ref: ${{ github.event.inputs.gitRef }}
     - name: Create release branch
+      if: ${{ !startsWith(github.event.inputs.gitRef, 'release-v') }}
       run: |
         git checkout -b release-v${{ github.event.inputs.operatorVersion }}
     - name: Prepare release
@@ -51,3 +56,4 @@ jobs:
         body: "**This release enables installations of Authorino v${{ github.event.inputs.authorinoVersion }}**"
         generate_release_notes: true
         target_commitish: release-v${{ github.event.inputs.operatorVersion }}
+        prerelease: ${{ github.event.inputs.prerelease }}
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -41,7 +41,7 @@ jobs:
       run: |
         git checkout -b release-v${{ github.event.inputs.operatorVersion }}
     - name: Prepare release
-      run: VERSION=${{ github.event.inputs.operatorVersion }} AUTHORINO_VERSION=${{ github.event.inputs.authorinoVersion }} GH_TOKEN=${{ secrets.GITHUB_TOKEN }} make prepare-release
+      run: VERSION=${{ github.event.inputs.operatorVersion }} AUTHORINO_VERSION=${{ github.event.inputs.authorinoVersion }} make prepare-release
     - name: Commit and push
       run: |
         git config --global user.name "github-actions[bot]"
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,7 @@ COPY pkg/ pkg/
 
 ARG VERSION=latest
 ARG DEFAULT_AUTHORINO_IMAGE=quay.io/kuadrant/authorino:latest
-RUN CGO_ENABLED=0 GO111MODULE=on go build -a -ldflags "-X main.version=${VERSION} -X controllers.defaultAuthorinoImage=${DEFAULT_AUTHORINO_IMAGE}" -o manager main.go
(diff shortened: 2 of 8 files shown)


.github/workflows/release.yaml
```diff
@@ -0,0 +1,53 @@
...
+      id: go
+    - name: Checkout code at commit SHA
+      uses: actions/checkout@v3
+      with:
+        ref: ${{ github.event.inputs.commitSha }}
+    - name: Create release branch
+      run: |
+        git checkout -b release-v${{ github.event.inputs.operatorVersion }}
+    - name: Prepare release
+      run: VERSION=${{ github.event.inputs.operatorVersion }} AUTHORINO_VERSION=${{ github.event.inputs.authorinoVersion }} GH_TOKEN=${{ secrets.GITHUB_TOKEN }} make prepare-release
```
guicassolato, line 39: We need to make sure the secret won't leak to the logs.

Makefile
```diff
@@ -214,22 +241,22 @@ bundle-build: ## Build the bundle image.
...
+               https://api.github.com/repos/Kuadrant/authorino-operator/releases/latest | \
+               jq -r '.name'))
+
+	@echo $(REPLACES_VERSION)
+	V="authorino-operator.$(REPLACES_VERSION)" $(YQ) eval '.spec.replaces = strenv(V)' -i bundle/manifests/authorino-operator.clusterserviceversion.yaml
+
+.PHONY: prepare-release
+prepare-release: 
+	$(MAKE) manifests bundle VERSION=$(VERSION) AUTHORINO_VERSION=$(AUTHORINO_VERSION)
+	sed -i 's|\(defaultAuthorinoImage\s\+string = "quay.io/kuadrant/authorino:\).*"|\1v'$(AUTHORINO_VERSION)'"|' controllers/constants.go
```
guicassolato, line 258: This is fine for a first iteration. For a subsequent improvement, I was thinking... The `controllers.defaultAuthorinoImage` constant could be set by using `-ldflags` when we `go build`. Of course, the value we set should come from somewhere. This could be file that we here create in the root of the project. `make build` (as well as `make run`) would pick up the file and set the flag. If the file is missing, it would default to `quay.io/kuadrant/authorino:latest`. WDYT?

.github/workflows/release.yaml
adam-cattermole: Whilst I can't find exact proof, I _think_ that this `GITHUB_TOKEN` secret should be masked in the runner. I can see an example usage here https://docs.github.com/en/actions/security-guides/automatic-token-authentication#example-2-calling-the-rest-api

.github/workflows/release.yaml
```diff
@@ -0,0 +1,53 @@
...
+      with:
+        go-version: 1.19.x
+      id: go
+    - name: Checkout code at commit SHA
+      uses: actions/checkout@v3
+      with:
+        ref: ${{ github.event.inputs.commitSha }}
+    - name: Create release branch
+      run: |
+        git checkout -b release-v${{ github.event.inputs.operatorVersion }}
```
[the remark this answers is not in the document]
adam-cattermole: Would this be to allow us to re-run on an existing `release-` branch?

Makefile
adam-cattermole: I definitely prefer this over `sed` actually. We would also have to check for the presence of this file in the Dockerfile where we build the binary too right (https://github.com/Kuadrant/authorino-operator/blob/main/Dockerfile#L20) to make sure that the correct version is set when the build and push images pipeline runs?

.github/workflows/release.yaml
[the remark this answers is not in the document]
guicassolato: Good question. IDK why we would release an already released branch. I was thinking on a release branch we create upfront, add changes that are only meant to that release (e.g. cherry-picking patches) and then make a release.

If we don't add the condition I suggested, it's still doable though. Create a temp release branch with whatever name, add the changes to this branch, make a release from this branch and then, if you want, delete the temp branch.

Makefile
```diff
@@ -214,22 +244,24 @@ bundle-build: ## Build the bundle image.
...
-OPM = $(shell which opm)
-endif
-endif
+.PHONY: fix-csv-replaces
+fix-csv-replaces: $(YQ)
+	$(eval REPLACES_VERSION=$(shell curl -sSL -H "Accept: application/vnd.github+json" \
+               -H "Authorization: Bearer ${GH_TOKEN}" \
+               -H "X-GitHub-Api-Version: 2022-11-28" \
+               https://api.github.com/repos/Kuadrant/authorino-operator/releases/latest | \
+               jq -r '.name'))
```
guicassolato, line 253: This also works without the `GH_TOKEN`, doesn't it?

```sh
❯ curl -sSL \
  -H "Accept: application/vnd.github+json" \
  -H "X-GitHub-Api-Version: 2022-11-28" \
  https://api.github.com/repos/Kuadrant/authorino-operator/releases/latest | jq -r '.name'
v0.8.0
```

adam-cattermole: Aha you're not wrong - we followed the docs described here which did not state the token as optional for public repos - https://docs.github.com/en/free-pro-team@latest/rest/releases/releases?apiVersion=2022-11-28#get-the-latest-release

If this is the case I'd prefer to opt for removing the `GH_TOKEN` altogether

guicassolato: > I'd prefer to opt for removing the `GH_TOKEN`

I was thinking the same. The main (only?) benefit of the token here is to avoid getting rate-limited, but since this is for the release workflow, it's not like it's something that requires many requests anyway.

adam-cattermole: Yep agreed, we'll update this

guicassolato (changes requested):
A couple extra things missing here. (Sorry. I've just spotted them now.)

1. The [`defaultAuthorinoImage`](https://github.com/Kuadrant/authorino-operator/blob/6bd81ee1757e2f565bc5a0a17140330cb222bce0/controllers/constants.go#L56) constant needs to become a variable or the ldflag will not work.

```diff
diff --git a/controllers/constants.go b/controllers/constants.go
index dd49876..f4ceb4f 100644
--- a/controllers/constants.go
+++ b/controllers/constants.go
@@ -53,7 +53,6 @@ const (
 	flagMaxHttpRequestBodySize         string = "max-http-request-body-size"
 
 	// defaults
-	defaultAuthorinoImage      string = "quay.io/kuadrant/authorino:latest"
 	defaultTlsCertPath         string = "/etc/ssl/certs/tls.crt"
 	defaultTlsCertKeyPath      string = "/etc/ssl/private/tls.key"
 	defaultOidcTlsCertPath     string = "/etc/ssl/certs/oidc.crt"
@@ -89,3 +88,6 @@ const (
 	statusUnableToUpdateDeployment                = "UnableToUpdateDeployment"
 	statusDeploymentNotReady                      = "DeploymentNotReady"
 )
+
+// ldflags
+var defaultAuthorinoImage string
```

2. We need to specify the full path of the variable to modify at build-time

```diff
diff --git a/Makefile b/Makefile
index 10ebdf0..1b58af0 100644
--- a/Makefile
+++ b/Makefile
@@ -167,10 +167,10 @@ test: manifests generate fmt vet setup-envtest
 ##@ Build
 
 build: generate fmt vet ## Build manager binary.
-	go build -ldflags "-X main.version=$(VERSION) -X controllers.defaultAuthorinoImage=$(DEFAULT_AUTHORINO_IMAGE)" -o bin/manager main.go
+	go build -ldflags "-X main.version=$(VERSION) -X github.com/kuadrant/authorino-operator/controllers.defaultAuthorinoImage=$(DEFAULT_AUTHORINO_IMAGE)" -o bin/manager main.go
 
 run: manifests generate fmt vet ## Run a controller from your host.
-	go run -ldflags "-X main.version=$(VERSION) -X controllers.defaultAuthorinoImage=$(DEFAULT_AUTHORINO_IMAGE)" ./main.go
+	go run -ldflags "-X main.version=$(VERSION) -X github.com/kuadrant/authorino-operator/controllers.defaultAuthorinoImage=$(DEFAULT_AUTHORINO_IMAGE)" ./main.go
 
```

BTW, to check the name of a variable you want to set using `-ldflags`, you can use the `go tool nm` command:

```
❯ go tool nm bin/manager| grep main.version
101dee240 D main.version
101088f00 R main.version.str

❯ go tool nm bin/manager| grep controllers.defaultAuthorinoImage
101e05140 B github.com/kuadrant/authorino-operator/controllers.defaultAuthorinoImage
```

guicassolato:
Tests are failing now because of missing a value for `controllers.defaultAuthorinoImage`; the Authorino deployments do not get a valid image and therefore are rejected. We may need to set the `-ldflags` in `go test` as well.

Makefile
grzpiotrowski: Just pushed the changes as suggested. However, in the meantime I noticed the workflow failing when authorinoVersion parameter is left as "latest". Works fine when it is set to a specific version like 0.14.0 for example.

Dockerfile
```diff
@@ -16,8 +16,9 @@ COPY api/ api/
 COPY controllers/ controllers/
 COPY pkg/ pkg/
 
-ARG version=latest
-RUN CGO_ENABLED=0 GO111MODULE=on go build -a -ldflags "-X main.version=${version}" -o manager main.go
+ARG VERSION=latest
+ARG DEFAULT_AUTHORINO_IMAGE=quay.io/kuadrant/authorino:latest
+RUN CGO_ENABLED=0 GO111MODULE=on go build -a -ldflags "-X main.version=${VERSION} -X controllers.defaultAuthorinoImage=${DEFAULT_AUTHORINO_IMAGE}" -o manager main.go
```
guicassolato, line 21:
```suggestion
RUN CGO_ENABLED=0 GO111MODULE=on go build -a -ldflags "-X main.version=${VERSION} -X github.com/kuadrant/authorino-operator/controllers.defaultAuthorinoImage=${DEFAULT_AUTHORINO_IMAGE}" -o manager main.go
```

guicassolato:
Now that we will no longer be able to read the default authorino image anywhere in the code, what do you think about logging it together with the operator version?

https://github.com/Kuadrant/authorino-operator/blob/254b656befd65e29fe55b9d8c0f746c254894d53/main.go#L70

PR #129, opened 2023-07-31
codecov: do not fail ci on error

KevFan:
* Prevent codecov from failing ci on error due to flakiness with codecov upload
* Use `CODECOV_TOKEN` for upload that should at least prevent upload issue when PRs come from branches from this repo


adam-cattermole (approved):
Changes look good to me

PR #130, opened 2023-08-07
Remove unused branch

adam-cattermole:
Minor change to remove an unused `if` branch. The `desiredDeployment` is created with a single container prior to this so I believe the check should just be removed.


KevFan (approved):
Looks good to me 👍

PR #131, opened 2023-08-08
Update to ginkgo v2

### LINKED ISSUE #126: dep: use ginkgo v2
* Ginkgo v1 is deprecated.  We should migrate to use v2 instead
  * https://onsi.github.io/ginkgo/MIGRATING_TO_V2

adam-cattermole:
Resolves #126


KevFan (approved):
Changes looks good to me 👍

PR #132, opened 2023-08-09
Set commit sha in bundle/catalog

### LINKED ISSUE #121: Include commit SHA for related images in the operator catalog
At the moment, in the task creating the catalog, there's only a distinction when the build is for `main` or a `release` branch, tagging the catalog related images as `latest` and the specific `vX.Y.Z` release tag... but there's no condition for the SHA tagged build.

The current latest SHA build is rendered as:

```sh
bin/opm render quay.io/kuadrant/authorino-operator-catalog:30fb663ddc5b2978add48fd6ee9ff526317c05a5 --output=yaml


...

---
entries:
- name: authorino-operator.v0.0.0
name: alpha
package: authorino-operator
schema: olm.channel
---
image: quay.io/kuadrant/authorino-operator-bundle:latest # This should be quay.io/kuadrant/authorino-operator-bundle:30fb663ddc5b2978add48fd6ee9ff526317c05a5
name: authorino-operator.v0.0.0
package: authorino-operator
properties:
...

relatedImages:
- image: quay.io/kuadrant/authorino-operator:latest  #This should be quay.io/kuadrant/authorino-operator:30fb663ddc5b2978add48fd6ee9ff526317c05a5

...
```

Putting aside that this operator should build the catalog with the new file system instead of the current deprecated Sqlite-based one, the proposed change would be to include the SHA commit as a tag in the GH action that creates the catalog in `.github/workflows/build-images.yaml` and possibly make changes in the invoked target in the `Makefile`

adam-cattermole:
Updated the default flow to set the commit sha in the bundle manifests and in the catalog. This makes it so that the `relatedImages` and the bundle in the catalog are pointing to the commit sha variant of the images. As the workflow pushes duplicate images for both the `latest` and commit sha, this means that the latest images will also be pointing to specific commit sha tagged images.

Due to the pipeline modifying the manifests to set the image to the commit sha variant, the `verify-manifests` and `verify-bundle` make targets will fail. These checks are verifying that there are no uncommitted changes in the local git repo which does not make sense given the changes, and so the check was removed for the `main` flow.

Verified this on our own branch [here](https://github.com/adam-cattermole/authorino-operator/actions/runs/5807542022). Example output from the render:

```
opm render quay.io/acatterm/authorino-operator-catalog:d350f6a0f8f4e705a3fc9ac5b048967479a6503a --output=yaml
```
```yaml
---
image: quay.io/acatterm/authorino-operator-bundle:d350f6a0f8f4e705a3fc9ac5b048967479a6503a
name: authorino-operator.v0.0.0
package: authorino-operator
...

relatedImages:
- image: quay.io/acatterm/authorino-operator:d350f6a0f8f4e705a3fc9ac5b048967479a6503a
  name: ""
schema: olm.bundle
```

Looking for thoughts on this approach.

Resolves #121

### DIFF
--- a/.github/workflows/build-images.yaml
+++ b/.github/workflows/build-images.yaml
@@ -115,7 +115,7 @@ jobs:
           sudo apt-get install -y qemu-user-static
       - name: Run make bundle (main)
         if: ${{ github.ref_name == env.MAIN_BRANCH_NAME }}
-        run: make bundle REGISTRY=${{ env.IMG_REGISTRY_HOST }} ORG=${{ env.IMG_REGISTRY_ORG }}
+        run: make bundle REGISTRY=${{ env.IMG_REGISTRY_HOST }} ORG=${{ env.IMG_REGISTRY_ORG }} IMAGE_TAG=${{ github.sha }}
       - name: Run make bundle (release)
         if: ${{ github.ref_name != env.MAIN_BRANCH_NAME }}
         run: make bundle REGISTRY=${{ env.IMG_REGISTRY_HOST }} ORG=${{ env.IMG_REGISTRY_ORG }} VERSION=${TAG_NAME/v/} AUTHORINO_VERSION=${{ github.event.inputs.authorinoVersion }}
@@ -123,7 +123,7 @@ jobs:
         run: git diff
       - name: Verify manifests and bundle (main)
         if: github.ref_name == env.MAIN_BRANCH_NAME
-        run: make verify-manifests verify-bundle REGISTRY=${{ env.IMG_REGISTRY_HOST }} ORG=${{ env.IMG_REGISTRY_ORG }}
+        run: make verify-manifests verify-bundle REGISTRY=${{ env.IMG_REGISTRY_HOST }} ORG=${{ env.IMG_REGISTRY_ORG }} IMAGE_TAG=${{ github.sha }}
       - name: Verify manifests and bundle (release)
         if: ${{ github.ref_name != env.MAIN_BRANCH_NAME }}
         run: make verify-manifests verify-bundle REGISTRY=${{ env.IMG_REGISTRY_HOST }} ORG=${{ env.IMG_REGISTRY_ORG }} VERSION=${TAG_NAME/v/} AUTHORINO_VERSION=${{ github.event.inputs.authorinoVersion }}
@@ -190,7 +190,7 @@ jobs:
           sudo apt-get install -y qemu-user-static
       - name: Run make catalog-generate (main)
         if: ${{ github.ref_name == env.MAIN_BRANCH_NAME }}
-        run: make catalog-generate REGISTRY=${{ env.IMG_REGISTRY_HOST }} ORG=${{ env.IMG_REGISTRY_ORG }}
+        run: make catalog-generate REGISTRY=${{ env.IMG_REGISTRY_HOST }} ORG=${{ env.IMG_REGISTRY_ORG }} IMAGE_TAG=${{ github.sha }}
       - name: Run make catalog-generate (release)
         if: ${{ github.ref_name != env.MAIN_BRANCH_NAME }}
         run: make catalog-generate REGISTRY=${{ env.IMG_REGISTRY_HOST }} ORG=${{ env.IMG_REGISTRY_ORG }} IMAGE_TAG=${{ env.TAG_NAME }}
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -23,6 +23,21 @@ jobs:
         run: |
           make verify-manifests
 
+  verify-bundle:
+    name: Verify bundle
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up Go 1.19.x
+        uses: actions/setup-go@v4
+        with:
+          go-version: 1.19.x
+        id: go
+      - name: Check out code
+        uses: actions/checkout@v3
+      - name: Run make verify-bundle
+        run: |
+          make verify-bundle
+
   verify-fmt:
     name: Verify fmt
     runs-on: ubuntu-latest
--- a/Makefile
+++ b/Makefile
@@ -296,12 +296,12 @@ catalog-push: ## Push a catalog image.
 
 .PHONY: verify-manifests
 verify-manifests: manifests ## Verify manifests update.
-	git diff --exit-code ./config
+	git diff -I'^    containerImage:' -I'^        image:' --exit-code ./config
 	[ -z "$$(git ls-files --other --exclude-standard --directory --no-empty-directory ./config)" ]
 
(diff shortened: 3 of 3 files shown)


.github/workflows/build-images.yaml
```diff
@@ -115,15 +115,12 @@ jobs:
           sudo apt-get install -y qemu-user-static
       - name: Run make bundle (main)
         if: ${{ github.ref_name == env.MAIN_BRANCH_NAME }}
-        run: make bundle REGISTRY=${{ env.IMG_REGISTRY_HOST }} ORG=${{ env.IMG_REGISTRY_ORG }}
+        run: make bundle REGISTRY=${{ env.IMG_REGISTRY_HOST }} ORG=${{ env.IMG_REGISTRY_ORG }} IMAGE_TAG=${{ github.sha }}
```
didierofrivia, line 118: We can remove the _SHA_ tag from main branch, since it's not needed and causing the failing bundle and manifest verification. With the branch ones it will do, since the job collecting the sha tags from quay only cares for the latest one:
https://github.com/Kuadrant/kuadrant-operator/blob/main/.github/workflows/update-stored-dependencies-version.yaml
Then the scheduled event will create kuadrant bundle and catalog based on those here:
https://github.com/Kuadrant/kuadrant-operator/blob/main/.github/workflows/build-images-scheduled.yaml

.github/workflows/build-images.yaml
```diff
@@ -115,15 +115,12 @@ jobs:
...
       - name: Run make bundle (release)
         if: ${{ github.ref_name != env.MAIN_BRANCH_NAME }}
         run: make bundle REGISTRY=${{ env.IMG_REGISTRY_HOST }} ORG=${{ env.IMG_REGISTRY_ORG }} VERSION=${TAG_NAME/v/} AUTHORINO_VERSION=${{ github.event.inputs.authorinoVersion }}
       - name: Git diff
         run: git diff
-      - name: Verify manifests and bundle (main)
```
didierofrivia, line 124: Then this step could be added back

didierofrivia (changes requested):
If with the `IMAGE_TAG` change is enough for the catalog generation, it's great... in the future we could update the catalog generation to file based instead of sqlite, to align with the others.. or just have one workflow to rule them all. Anyways, the fork kind of breaks the implementation, because it can't push to quay, then the scheduled job will always be pulling a stale version...

.github/workflows/build-images.yaml
adam-cattermole: I'm not quite sure I follow here - if we remove the `IMAGE_TAG` from the `main` flow all catalog images will contain the bundle tagged as `latest` and not the specific commit sha.

didierofrivia:
We will need to change the way of generating this in the end, more information can be found in [this issue](https://github.com/Kuadrant/kuadrant-operator/issues/227)

.github/workflows/build-images.yaml
didierofrivia: In the end, the sha will need to be for main only, and fix the verification stage. We were relying on the generation of the latest sha pushed to quay (from branches) to generate the kuadrant-operator catalog in a scheduled way for automated testing needed for QE (the workflows shared in my first comment). After some thought, given the restrictions on the before mentioned implementation, it needs to be changed to only main. This would be the main issue https://github.com/Kuadrant/kuadrant-operator/issues/227.

didierofrivia: Sorry for the confusion generated and the lack of information in the original issue.

.github/workflows/build-images.yaml
guicassolato: Regardless, I would not remove this check. It ensures (i) that the bundle generation is idempotent, and (ii) that we did not forget to commit any important change in the copy of the manifests that is pushed to remote.

Hypothetically – Ignoring what @didierofrivia [said](https://github.com/Kuadrant/authorino-operator/pull/132/files#r1289908207) for a second, if we were to include the SHA in the bundle manifests (like proposed here), then, to prevent from failing at this step (since that SHA would not match what I imagine would have been committed and pushed to remote, i.e. `latest`), the verification of the bundle must be enhanced.

adam-cattermole: > Regardless, I would not remove this check. It ensures (i) that the bundle generation is idempotent, and (ii) that we did not forget to commit any important change in the copy of the manifests that is pushed to remote.

Just to note the [verify-manifests](https://github.com/Kuadrant/authorino-operator/blob/main/.github/workflows/test.yaml#L11) check that runs on PR's is unchanged - the check we removed above only runs when commits are merged to main (which should have already passed the verify-manifests check) or on `workflow_dispatch`. We could add to the test workflow the `make verify-bundle` command so that PR's can't be missing the changes in the manifests/bundles, but this would still mean that we are making the assumption that the main branch is always in a valid state as a result of this.

adam-cattermole:
After discussion we have added the check to `make verify-manifests verify-bundle` back into the workflow and ignored the lines that specify the images with the commit sha. Also added an additional commit where we add the `make verify-bundle` to the test workflow run on PR's.

guicassolato (approved):
_LGTM_

I see this is as the minimal change to get the issue of the floating `latest` image tag fixed. This has been causing problems to @Kuadrant/qe who need to lock the environment to a particular set of SHAs in order to debug stuff and relate the issues they found to specific versions of the code.

<br/>

Maybe future enhancements could include:
- **A.** Somehow avoid that the image tag that is injected in the bundle/catalog files by the workflow (based on the commit SHA) mismatches what's read in the code by one who checks out the main branch (i.e. `:latest`).[^1]
- **B.** Standardise how the Authorino Operator generates the bundle manifests and catalog image compared to other operators ([kuadrant/kuadrant-operator](https://github.com/kuadrant/kuadrant-operator), [kuadrant/limitador-operator](https://github.com/kuadrant/limitador-operator)) – e.g., see https://github.com/Kuadrant/limitador-operator/pull/62

[^1]: I few different ways to try to achieve that (mentioned in an off-line call we had about this change): (i) some kind of post-commit hook that sets the SHA when the bundle manifests are generated locally, before pushing them to remote; (ii) replacing the commit SHA with another system to lock the manifests to a fixed version of the code (e.g. timestamp-based); (iii) start git-ignoring the bundle manifests altogether and start generating them on-the-fly whenever we need to build and push the image/cut a release only; (iv) get back to using the floating tag `latest` and create a special workflow for QE that follows the same steps as when we build for a release, i.e. introduce automated pre-releases from the main branch only for QE (after very merge, daily, weekly, etc).

pehala (approved):

adam-cattermole:
@didierofrivia If you have time do you mind reviewing the latest changes here to make some progress on this?

didierofrivia (approved):
🎖️

PR #134, opened 2023-08-21
Namespace support.

### LINKED ISSUE #41: Possibility to deploy the operator in an arbitrary namespace
Authorino Operator's [deployment manifests](https://github.com/Kuadrant/authorino-operator/blob/main/config/deploy/manifests.yaml), generated using [this](https://github.com/Kuadrant/authorino-operator/blob/785b0777ac031fbd2a10cfa5478831b8db9419b2/Makefile#L171) make target, currently enforce `authorino-operator` as the Kubernetes namespace to run the operator workload.

While on one hand that single-file manifest is very useful to install and deploy the Operator and the CRDs, changing the operator namespace, on the other hand, is not straightforward, requiring all occurrences of `namespace: authorino-operator` in the file, as well as the name of `Namespace` resource itself, to be properly replaced.

Except in the `RoleBinding` and `ClusterRoleBinding` resources, mentioning the Kubernetes namespace could be avoided, but in general, perhaps using a placeholder to ensure assertive replacement (e.g. making possible to rely on envsubst)? Another option would be relying more on [Operator Lifecyle Manager](https://olm.operatorframework.io) to install the operator? Although it's not available by default in Kubernetes, but only in OpenShift.

Boomatang:
- Add support for setting the namespace during the manifest creations
- Add make targets for creation of kind cluster for local development

Closes #41

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -104,6 +104,10 @@ KUSTOMIZE = $(shell pwd)/bin/kustomize
 kustomize: ## Download kustomize locally if necessary.
 	$(call go-get-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v4@v4.5.5)
 
+KIND = $(shell pwd)/bin/kind
+kind:
+	$(call go-get-tool,$(KIND),sigs.k8s.io/kind@v0.20.0)
+
 YQ = $(shell pwd)/bin/yq
 YQ_VERSION := v4.34.2
 $(YQ):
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -307,3 +345,72 @@ verify-bundle: bundle ## Verify bundle update.
 .PHONY: verify-fmt
 verify-fmt: fmt ## Verify fmt update.
 	git diff --exit-code ./api ./controllers
+
+## local configurations
+
+deploy-develmode: manifests kustomize ## Deploy controller in debug mode to the K8s cluster specified in ~/.kube/config.
```
guicassolato, line 351: The target name says "devel" mode while the descriptions says "debug". But, most importantly:
- What's the difference to the existing `deploy` target?
- Why does it override the deployment manifests file at `$(DEPLOYMENT_FILE)`, but then won't use it?

Makefile
```diff
@@ -182,16 +186,40 @@ docker-push: ## Push docker image with the manager.
 ##@ Deployment
 
 install: manifests kustomize install-authorino ## Install CRDs into the K8s cluster specified in ~/.kube/config.
+	@if [ $(NAMESPACE) != '' ];then \
+		echo "Setting Custom Namespace: $(NAMESPACE)"; \
+		cd $(PROJECT_DIR)/config/install && $(KUSTOMIZE) edit set namespace $(NAMESPACE); \
```
guicassolato, line 191: Do you have the diff of what this target yields with a custom namespace vs the default one? Likely, this is only because of the role bindings created for the operator, so the it sets the right ServiceAccount, correct?

Makefile
```diff
@@ -307,3 +345,72 @@ verify-bundle: bundle ## Verify bundle update.
...
+
+KIND_CLUSTER_NAME ?= authorino-local
+
+.PHONY: kind-create-cluster
+kind-create-cluster: kind ## Create the "authorino-local" kind cluster.
+	$(KIND) create cluster --name $(KIND_CLUSTER_NAME) --config utils/kind-cluster.yaml
+
+.PHONY: kind-delete-cluster
+kind-delete-cluster: kind ## Delete the "authorino-local" kind cluster.
+	$(KIND) delete cluster --name $(KIND_CLUSTER_NAME)
```
guicassolato, line 416: I'm guessing these local cluster-related targets are all to support testing the operator deployed to a custom namespace. Was that the initial motivation?

To test local changes in the operator, I often find myself running the following:

```sh
kind create cluster
kind load docker-image authorino-operator:local
make manifests && sed -e 's/quay.io\/kuadrant\/authorino-operator:latest/authorino-operator:local/' config/deploy/manifests.yaml | kubectl apply -f -
```

Some other times:

```sh
kind create cluster
make manifests install run
```

Makefile
```diff
@@ -307,3 +345,72 @@ verify-bundle: bundle ## Verify bundle update.
...
+	$(MAKE) local-env-setup
+	$(MAKE) docker-build
+	echo "Deploying Authorino control plane"
+	$(KIND) load docker-image ${OPERATOR_IMAGE} --name ${KIND_CLUSTER_NAME}
+	$(MAKE) install
+	$(MAKE) deploy
+
+.PHONY: local-redeploy
+local-redeploy: export OPERATOR_IMAGE := authorino-operator:dev
+local-redeploy:
```
guicassolato, line 391: For consistency with how Authorino [calls](https://github.com/Kuadrant/authorino/blob/374cd7c42f80babc84e9f960be2bd9d8e63fde41/Makefile#L261) it:

```suggestion
local-rollout:
```

Makefile
Boomatang: My initial reason for adding  the local cluster related targets was for similarity in the dev process. Both authorino and kuadrant-operator uses the same naming. The steps you stated are fine but they are not documented so when starting in the project it is unclear how to start with a dev cluster.

Makefile
Boomatang: I'll create a diff

Makefile
Boomatang: At the I am going to say I had a valid reason for doing it but I can not find it now. I will remove and use the `deploy` target.

guicassolato:
In general, I think the changes here are welcome for the developer workflow, i.e., the one where you clone the repo, make local changes, and want to try them out by rebuilding and deploying to an almost always local cluster, whilst switching the namespace to something other than the default `authorino-operator`.

For the non-OLM user workflow nonetheless, i.e., the one by which the operator is [installed by simply applying the pre-build manifests](https://github.com/Kuadrant/authorino/blob/main/docs/user-guides/hello-world.md#install-the-authorino-operator) (without cloning the repo, without custom build of the binary or container image), it's a step forward, but likely won't suffice. For that, we may need to incorporate some of the changes here into the install script to be introduced with https://github.com/Kuadrant/authorino-operator/pull/137.

I'm thinking a possible subsequent step after merging both PRs would be making the install script to rely as well on kustomize, to rebuild the manifests "on-the-fly" including an optional custom namespace. MGC team did sth similar with https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/hack/quickstart-setup.sh.

Makefile
Boomatang: The diff is in the gist. I comment out then clean up code when creating the diff.
https://gist.github.com/Boomatang/becfda6a1659f7e864ac039a439889bd

PR #137, opened 2023-08-29
AuthConfig conversion webhook

guicassolato:
Deploys the AuthConfig CRD conversion webhook (based on the Authorino container image), as part of deploying the Operator.

This is because the conversion webhook is a single deployment per cluster (similarly to the Operator itself), nevertheless it's based on the Authorino code base (which owns the AuthConfig type and therefore the functions to convert between versions of the CRD).

**_Important!_**

This change introduces a new dependency of the Operator on [cert-manager](https://cert-manager.io). Actually the dependency is introduced by the conversion webhook (not by the Operator itself), but since installing/deploying the Operator will also deploy the webhook, cert-manager needs to be installed before installing the Operator.

Because of that, installing the Operator by simply applying the `config/deploy` manifests may no longer work, with the webhook possibly failing due to cert-manager not running in the cluster. Hence, the PR also introduces a script to continue supporting installing the Operator – and now also cert-manager –  without having to clone the repo nor depending on Operator Lifecycle Manager (OLM).

The script can be used as follows:

```sh
curl -sL https://raw.githubusercontent.com/Kuadrant/authorino-operator/main/utils/install.sh | bash -s -- [options]
```

The options include : `--version`, `--git-ref`, `--no-deps`, and `--dry-run`.

For installations via OLM, installing cert-manager first is [not required](https://olm.operatorframework.io/docs/advanced-tasks/adding-admission-and-conversion-webhooks/).

---

Related to https://github.com/Kuadrant/authorino/pull/417

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -137,6 +137,7 @@ manifests: controller-gen kustomize authorino-manifests ## Generate WebhookConfi
 
 .PHONY: authorino-manifests
 authorino-manifests: export AUTHORINO_GITREF := $(AUTHORINO_BRANCH)
+authorino-manifests: export AUTHORINO_VERSION := $(AUTHORINO_VERSION)
 authorino-manifests: ## Update authorino manifests.
 	envsubst \
         < config/authorino/kustomization.template.yaml \
--- a/bundle/manifests/authorino-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/authorino-operator.clusterserviceversion.yaml
@@ -43,6 +43,9 @@ spec:
       kind: AuthConfig
       name: authconfigs.authorino.kuadrant.io
       version: v1beta1
+    - kind: AuthConfig
+      name: authconfigs.authorino.kuadrant.io
+      version: v1beta2
     - description: API to create instances of authorino
       displayName: Authorino
       kind: Authorino
--- a/bundle/manifests/authorino-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/authorino-operator.clusterserviceversion.yaml
@@ -280,6 +280,45 @@ spec:
                 runAsNonRoot: true
               serviceAccountName: authorino-operator
               terminationGracePeriodSeconds: 10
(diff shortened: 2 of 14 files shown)


guicassolato:
The **Verify manifests** test failing is expected, because https://github.com/Kuadrant/authorino/pull/417 has not been merged yet.

Boomatang (dismissed):
Running the script worked as expected. By using the `--git-ref` flag I a was able to install from this branch and all the web hook got deployed as expected.

adam-cattermole (approved):
I verified these changes and the script standalone (the creation/deletion of authconfig v1beta2 with the webhook server), as well as alongside the changes in authorino through running the user guides and everything looks good to me

didierofrivia (approved):
Works as expected! using the script provided in this branch with `--git-ref conversion-webhook `

didierofrivia:
Ok, something I've just seen when installing it with `curl -sL https://raw.githubusercontent.com/Kuadrant/authorino-operator/152a1627a0b2f5d36c3983192644ec8c5d625bad/utils/install.sh | bash -s -- --git-ref conversion-webhook --version authconfig-v1beta2` after waiting for a while, the webhook is still _progressing_

```sh
 k describe deployment/authorino-webhooks -n authorino-operator

...

Conditions:
  Type           Status  Reason
  ----           ------  ------
  Progressing    True    NewReplicaSetAvailable
  Available      False   MinimumReplicasUnavailable
OldReplicaSets:  <none>
NewReplicaSet:   authorino-webhooks-f77ddcc88 (1/1 replicas created)
Events:
  Type    Reason             Age    From                   Message
  ----    ------             ----   ----                   -------
  Normal  ScalingReplicaSet  4m52s  deployment-controller  Scaled up replica set authorino-webhooks-f77ddcc88 to 1
```

didierofrivia:
Right! thanks @adam-cattermole , I totally misread the script annotation xD

PR #139, opened 2023-09-22
Support local cert manager

Boomatang:
Add the installation of the cert manager for the tools used during dev work.

# Validation

* Create a local kind cluster
```
kind create cluster
```
* Set up the operator namespace
```
```
* Install the Operator manifests, cert manager will be add during this stage
```
make install
```
* Deploy the operator
```
make deploy
```
* Expect: Operator web hook to deploy. 
```
kubectl get deployment authorino-webhooks -n authorino-operator
```

Closes: https://github.com/Kuadrant/kuadrant-operator/issues/257

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,6 +58,9 @@ CRD_OPTIONS ?= "crd:trivialVersions=true,preserveUnknownFields=false"
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION = 1.21
 
+# Cert manager is required for the webhooks.
+CERT_MANAGER_VERSION ?= 1.12.1
+
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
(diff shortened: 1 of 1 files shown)


eguzki (approved):

Makefile
```diff
@@ -182,7 +185,7 @@ docker-push: ## Push docker image with the manager.
 
 ##@ Deployment
 
-install: manifests kustomize install-authorino ## Install CRDs into the K8s cluster specified in ~/.kube/config.
+install: manifests kustomize install-cert-manager install-authorino ## Install CRDs into the K8s cluster specified in ~/.kube/config.
```
guicassolato, line 188: Technically, cert-manager is a dependency for the Authorino webhook, which is deployed by `install-authorino`. I'd move it there, instead of `make install`.

guicassolato (commented):
Left a comment about which make target depends on `install-cert-manager`. Other than that, verification steps work, including `make run` alternatively to `make deploy`.

Makefile
Boomatang: Sure I am not going to argue with that.

guicassolato (approved):
Thank you!

PR #146, opened 2023-10-11
Disable bundle verify for release v0.9.0

adam-cattermole:
A bug in the bundle manifests verification is causing the builds to fail as the `replaces` directive is generated from the latest github release. During the release workflow the bundle manifests were created correctly (prior to publishing the github release) and have been verified manually. This PR is to disable these checks (on the release branch only) to build the bundle and catalog from the previously generated bundle files.


guicassolato (approved):
We all missed that when the automation for the release process was first introduced, @adam-cattermole. Thanks for taking care of it and thus making sure v0.9.0 is pushed outside the door.

Moreover and TBH, the verification of the manifests and bundle manifests has always been glitchy. On the flip side, when it works, it allows us to spot important changes often left behind, such as:
- updates to the `Authorino` API that were not reflected in the CRD;
- modifications to the base manifests without a counterpart to the bundle ones;
- updates in the manifests owned by [kuadrant/authorino](https://github.com/kuadrant/authorino) not brought together with the install and bundle manifests hosted in the operator repo.

Let's follow up on this by fixing the bug, and maybe later reviewing these workflows for carefully.

PR #156, opened 2023-11-07
Bump Go to 1.20

guicassolato:


adam-cattermole (approved):
Changes look good to me. Image builds + deploys no problem - ran through a user-guide for a quick test and all is fine

PR #158, opened 2023-11-08
Setup Authorino logger

### LINKED ISSUE #5: Migrate to use the same logging system as Authorino
Migrate the logging package to use the same as the once used in [authorino](https://github.com/Kuadrant/authorino/tree/main/pkg/common/log) 

See this [PR](https://github.com/Kuadrant/authorino/pull/172) for reference.

grzpiotrowski:
This change ensures that the authorino-operator uses _almost_ the same logger as in Authorino repo.

Only a small change made to the logger compared to the logger package in authorino, that is: in [SetLogger](https://github.com/Kuadrant/authorino/blob/main/pkg/log/logger.go#L82) function, getting rid of the Options parameter passed into it which was only used to log the level/mode inside that function.

Closes #5 

# Verification:

```
kind create cluster --name authorino-sandbox
```

Create the namespace for the Operator:
```
```

Build the local image:
```
docker build --build-arg DEFAULT_AUTHORINO_IMAGE=quay.io/kuadrant/authorino:latest -t authorino-operator:local .
```
Load the image into kind cluster:
```
kind -n authorino-sandbox load docker-image authorino-operator:local
```

Install operator manifests:
```
make install
```
Deploy authorino operator:
```
make deploy OPERATOR_IMAGE=authorino-operator:local
```

Deploy Authorino to have something to log:
```
echo "
kind: Authorino
metadata:
  name: authorino
spec:
  listener:
    tls:
      enabled: false
  oidcServer:
    tls:
      enabled: false
" | kubectl apply -f -
```

Check the logs in the default level and mode:
```
kubectl logs -n authorino-operator authorino-operator-XXXX
```

Change the LOG_LEVEL and LOG_MODE env variables:
```
kubectl patch deployment authorino-operator -n authorino-operator --type='json' -p='[
  {
    "op": "replace",
    "path": "/spec/template/spec/containers/0/args",
    "value": [
      "--leader-elect",
      "--log-level=debug",
      "--log-mode=development"
    ]
  }
]'
```

Check the logs of the new pod to see that different log level works:
```
kubectl logs -n authorino-operator authorino-operator-YYYY
```

Clean up:
```
kind delete cluster -n authorino-sandbox
docker image rm authorino-operator:local
```


main.go
```diff
@@ -39,15 +42,18 @@ import (
 var (
 	scheme   = runtime.NewScheme()
 	setupLog = ctrl.Log.WithName("setup")
-
-	version string // value injected in compilation-time
+	logLevel = common.FetchEnv("LOG_LEVEL", "info")
+	logMode  = common.FetchEnv("LOG_MODE", "production")
```
guicassolato, line 46: We've been moving away from env vars to configure the components, to favour command-line args instead.

Authorino still supports some env vars, including ones to configure the logging system, but that's only for backward compatibility. In all the cases, command-line args take precedence.

- https://github.com/Kuadrant/authorino/blob/8addd6752d9d424d295a70f6d75835069fdce0a6/main.go#L213-L214

In the Operator, I think even more so we should avoid introducing new env vars (because currently we have none.) Would you mind switching to command-line args instead please?

guicassolato (changes requested):

config/manager/manager.yaml
```diff
@@ -29,6 +29,8 @@ spec:
         - /manager
         args:
         - --leader-elect
+        - --log-level=info
+        - --log-mode=production
```
guicassolato, line 33: Since these are the defaults, we probably don't need them here. WDYT?

PR #161, opened 2023-11-23
Fix bugs in release workflow

adam-cattermole:


.github/workflows/build-images.yaml
```diff
@@ -122,15 +122,15 @@ jobs:
         run: make bundle REGISTRY=${{ env.IMG_REGISTRY_HOST }} ORG=${{ env.IMG_REGISTRY_ORG }} IMAGE_TAG=${{ github.sha }}
       - name: Run make bundle (release)
         if: ${{ github.ref_name != env.MAIN_BRANCH_NAME }}
-        run: make bundle fix-csv-replaces REGISTRY=${{ env.IMG_REGISTRY_HOST }} ORG=${{ env.IMG_REGISTRY_ORG }} VERSION=${TAG_NAME/v/} AUTHORINO_VERSION=${{ github.event.inputs.authorinoVersion }} CHANNELS=${{ github.event.inputs.channels }}
+        run: make bundle fix-csv-replaces REGISTRY=${{ env.IMG_REGISTRY_HOST }} ORG=${{ env.IMG_REGISTRY_ORG }} VERSION=${TAG_NAME/v/} AUTHORINO_VERSION=${{ github.event.inputs.authorinoVersion }} CHANNELS=${{ github.event.inputs.channels }} DEFAULT_CHANNEL=${{ github.event.inputs.channels }}
```
guicassolato, line 125: Channels can be a list IIRC, whereas the default one has to be, well, one. Right? Won't that be a problem?

.github/workflows/test.yaml
```diff
@@ -34,9 +34,9 @@ jobs:
         id: go
       - name: Check out code
         uses: actions/checkout@v3
-      - name: Run make verify-bundle
+      - name: Run make bundle verify-bundle
```
guicassolato, line 37:
```suggestion
      - name: Verify the bundle
```

.github/workflows/build-images.yaml
adam-cattermole: Yep you're right, I forgot we're not using file-based catalogs here but sqlite generated from bundle.

Makefile
```diff
@@ -312,7 +312,7 @@ verify-manifests: manifests $(YQ) ## Verify manifests update.
 	yq e -e '.metadata.annotations.containerImage == "$(OPERATOR_IMAGE)"' config/manifests/bases/authorino-operator.clusterserviceversion.yaml
 
 .PHONY: verify-bundle
-verify-bundle: bundle $(YQ) ## Verify bundle update.
+verify-bundle: $(YQ) ## Verify bundle update.
```
guicassolato, line 315: This virtually makes running `make verify-bundle` locally almost completely useless, but that's OK. In the end, locally, one can always run `make bundle` and thus ensure that the manifests are correct and up to date. Because, otherwise, they have to be committed and included with the changes.

`verify-bundle` only checks if the latest _committed_ bundle manifests need update, by re-running `make bundle` and checking if that leaves any unstaged diff. Perhaps this makes more sense in remote than to simply check locally if you didn't forget running `make bundle` before pushing your changes.

So let's make this target indicated for usage by the automated workflows only and not so much by devs locally.

.github/workflows/build-images.yaml
adam-cattermole: I've set this explicitly to `stable` for now as we only set the default channel in the release branches and it is currently assumed to be stable.

.github/workflows/test.yaml
adam-cattermole: Done

Makefile
```diff
@@ -83,8 +83,8 @@ AUTHORINO_BRANCH = v$(AUTHORINO_VERSION)
 AUTHORINO_IMAGE_TAG = v$(AUTHORINO_VERSION)
 endif
 
-AUTHORINO_IMAGE_FILE ?= authorino_image
-DEFAULT_AUTHORINO_IMAGE ?= $(shell cat $(AUTHORINO_IMAGE_FILE) || echo $(DEFAULT_REGISTRY)/$(DEFAULT_ORG)/authorino:latest)
+BUNDLE_CONFIG_FILE ?= bundle-config.yaml
```
guicassolato, line 86:
```suggestion
BUNDLE_CONFIG_FILE ?= config.yaml
```

or 

```suggestion
BUNDLE_CONFIG_FILE ?= build.yaml
```

Makefile
```diff
@@ -83,8 +83,8 @@ AUTHORINO_BRANCH = v$(AUTHORINO_VERSION)
...
+DEFAULT_AUTHORINO_IMAGE ?= $(shell yq e -e '.config.authorino-image' $(BUNDLE_CONFIG_FILE) || echo $(DEFAULT_REGISTRY)/$(DEFAULT_ORG)/authorino:latest)
```
guicassolato, line 87:
```suggestion
DEFAULT_AUTHORINO_IMAGE ?= $(shell yq e -e '.config.authorinoImage' $(BUNDLE_CONFIG_FILE) || echo $(DEFAULT_REGISTRY)/$(DEFAULT_ORG)/authorino:latest)
```

Makefile
```diff
@@ -255,22 +256,20 @@ bundle-build: ## Build the bundle image.
...
 .PHONY: prepare-release
 prepare-release:
+	$(MAKE) set-replaces-file
 	$(MAKE) manifests bundle VERSION=$(VERSION) AUTHORINO_VERSION=$(AUTHORINO_VERSION)
-	@if [ "$(AUTHORINO_VERSION)" = "latest" ]; then\
-		[ ! -e "$(AUTHORINO_IMAGE_FILE)" ] || rm $(AUTHORINO_IMAGE_FILE); \
-	else \
-	    echo quay.io/kuadrant/authorino:$(AUTHORINO_IMAGE_TAG) > $(AUTHORINO_IMAGE_FILE); \
+	@if [ "$(AUTHORINO_VERSION)" != "latest" ]; then\
+		V="$(DEFAULT_REGISTRY)/$(DEFAULT_ORG)/authorino:$(AUTHORINO_IMAGE_TAG)" $(YQ) eval '.config.authorino-image = strenv(V)' -i $(BUNDLE_CONFIG_FILE); \
```
guicassolato, line 271:
```suggestion
		V="$(DEFAULT_REGISTRY)/$(DEFAULT_ORG)/authorino:$(AUTHORINO_IMAGE_TAG)" $(YQ) eval '.config.authorinoImage = strenv(V)' -i $(BUNDLE_CONFIG_FILE); \
```

guicassolato (commented):
This looks great, @adam-cattermole! Thank you so much!

I've left a couple minor suggestions.

Makefile
```diff
@@ -255,22 +256,20 @@ bundle-build: ## Build the bundle image.
...
+set-replaces-file: $(YQ)
 	$(eval REPLACES_VERSION=$(shell curl -sSL -H "Accept: application/vnd.github+json" \
                https://api.github.com/repos/Kuadrant/authorino-operator/releases/latest | \
                jq -r '.name'))
-	V="authorino-operator.$(REPLACES_VERSION)" $(YQ) eval '.spec.replaces = strenv(V)' -i bundle/manifests/authorino-operator.clusterserviceversion.yaml
+	V="authorino-operator.$(REPLACES_VERSION)" $(YQ) -n '.config.replaces = strenv(V)' > $(BUNDLE_CONFIG_FILE)
 
 .PHONY: prepare-release
 prepare-release:
+	$(MAKE) set-replaces-file
```
guicassolato, line 268: Not sure if I like that the config file is only created when we fetch the latest release. What if I want one only for the default Authorino image option?

WDYT about:

1. `$(MAKE) create-config-file` (or `$(MAKE) create-build-file`, depending on what you choose at https://github.com/Kuadrant/authorino-operator/pull/161/files#r1403594058)
  1.1. `@echo "config:" > $(BUNDLE_CONFIG_FILE)`
  1.2. `$(MAKE) set-authorino-default-image`
  1.3. `$(MAKE) set-replaces-directive`
2. `$(MAKE) manifests bundle …`

?

adam-cattermole: Great suggestions, made this change

Makefile
```diff
@@ -312,7 +320,7 @@ verify-manifests: manifests $(YQ) ## Verify manifests update.
 	yq e -e '.metadata.annotations.containerImage == "$(OPERATOR_IMAGE)"' config/manifests/bases/authorino-operator.clusterserviceversion.yaml
 
 .PHONY: verify-bundle
-verify-bundle: bundle $(YQ) ## Verify bundle update.
+verify-bundle: $(YQ) ## Verify bundle update.
 	git diff -I'^    containerImage:' -I'^                image:' --exit-code ./bundle
 	[ -z "$$(git ls-files --other --exclude-standard --directory --no-empty-directory ./bundle)" ]
 	yq e -e '.metadata.annotations.containerImage == "$(OPERATOR_IMAGE)"' bundle/manifests/authorino-operator.clusterserviceversion.yaml
```
guicassolato, line 326: We should probably include here as well:

```
	yq e -e '.spec.install.spec.deployments[1].spec.template.spec.containers[0].image == "$(DEFAULT_AUTHORINO_IMAGE)"' bundle/manifests/authorino-operator.clusterserviceversion.yaml
	yq e -e '.spec.replaces == "$(shell yq e -e '.config.replaces' $(BUILD_CONFIG_FILE))"' bundle/manifests/authorino-operator.clusterserviceversion.yaml
```

Makefile
```diff
@@ -312,7 +320,7 @@ verify-manifests: manifests $(YQ) ## Verify manifests update.
 	yq e -e '.metadata.annotations.containerImage == "$(OPERATOR_IMAGE)"' config/manifests/bases/authorino-operator.clusterserviceversion.yaml
```
guicassolato, line 320: Because `bundle/manifests/authorino-operator.clusterserviceversion.yaml` is referred a few times in the Makefile, wdyt about declaring a constant for it?

Makefile
adam-cattermole: Won't the git diff catch these ones? the other ones for OPERATOR_IMAGE are because we explicitly ignore them in the git diff and verify with yq instead. Given the replaces + authorino image are set by the bundle run just prior to this the git diff should catch right?

Makefile
adam-cattermole: Yep I like this suggestion

Makefile
```diff
@@ -243,6 +247,7 @@ bundle: manifests kustomize operator-sdk ## Generate bundle manifests and metada
         < config/manifests/bases/authorino-operator.clusterserviceversion.template.yaml \
         > config/manifests/bases/authorino-operator.clusterserviceversion.yaml
 	$(KUSTOMIZE) build config/manifests | $(OPERATOR_SDK) generate bundle -q --overwrite --version $(BUNDLE_VERSION) $(BUNDLE_METADATA_OPTS) --package authorino-operator
+	($(YQ) e -e '.config.replaces' $(BUILD_CONFIG_FILE) && V="$(shell $(YQ) e -e '.config.replaces' $(BUILD_CONFIG_FILE))" $(YQ) eval '.spec.replaces = strenv(V)' -i $(BUNDLE_CSV)) || echo "no replaces added"
```
guicassolato, line 250:
```suggestion
	($(YQ) e -e '.config.replaces' $(BUILD_CONFIG_FILE) && V="$(shell $(YQ) e -e '.config.replaces' $(BUILD_CONFIG_FILE))" $(YQ) eval '.spec.replaces = strenv(V)' -i bundle/manifests/authorino-operator.clusterserviceversion.yaml) || ($(YQ) eval '.' -i bundle/manifests/authorino-operator.clusterserviceversion.yaml && echo "no replaces added")
```

So we make sure it's equally indented by default and therefore comparing the diff to `main` can be focused on what matters.

guicassolato:
### Verification steps

Prepare the release:

```sh
make prepare-release VERSION=0.10.0 AUTHORINO_VERSION=0.16.0 CHANNELS=stable DEFAULT_CHANNEL=stable
```

Check the manifests are correct:

```sh
git diff
```
```diff
diff --git a/bundle.Dockerfile b/bundle.Dockerfile
index 8fe9a64b..289359d2 100644
--- a/bundle.Dockerfile
+++ b/bundle.Dockerfile
@@ -5,7 +5,8 @@ LABEL operators.operatorframework.io.bundle.mediatype.v1=registry+v1
 LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=authorino-operator
-LABEL operators.operatorframework.io.bundle.channels.v1=alpha
+LABEL operators.operatorframework.io.bundle.channels.v1=stable
+LABEL operators.operatorframework.io.bundle.channel.default.v1=stable
 LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.22.0
 LABEL operators.operatorframework.io.metrics.mediatype.v1=metrics+v1
 LABEL operators.operatorframework.io.metrics.project_layout=go.kubebuilder.io/v3
diff --git a/bundle/manifests/authorino-operator.clusterserviceversion.yaml b/bundle/manifests/authorino-operator.clusterserviceversion.yaml
index d8cf8f64..e5db89e1 100644
--- a/bundle/manifests/authorino-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/authorino-operator.clusterserviceversion.yaml
@@ -26,13 +26,13 @@ metadata:
       ]
     capabilities: Basic Install
     categories: Integration & Delivery
-    containerImage: quay.io/kuadrant/authorino-operator:latest
+    containerImage: quay.io/kuadrant/authorino-operator:v0.10.0
     createdAt: 2021-12-08T10-00-00Z
     operators.operatorframework.io/builder: operator-sdk-v1.22.0
     operators.operatorframework.io/project_layout: go.kubebuilder.io/v3
     repository: https://github.com/Kuadrant/authorino-operator
     support: kuadrant
-  name: authorino-operator.v0.0.0
+  name: authorino-operator.v0.10.0
   namespace: placeholder
 spec:
   apiservicedefinitions: {}
@@ -256,7 +256,7 @@ spec:
                       - --leader-elect
                     command:
                       - /manager
-                    image: quay.io/kuadrant/authorino-operator:latest
+                    image: quay.io/kuadrant/authorino-operator:v0.10.0
                     livenessProbe:
                       httpGet:
                         path: /healthz
@@ -303,7 +303,7 @@ spec:
                   - command:
                       - authorino
                       - webhooks
-                    image: quay.io/kuadrant/authorino:latest
+                    image: quay.io/kuadrant/authorino:v0.16.0
                     name: webhooks
                     ports:
                       - containerPort: 9443
@@ -429,7 +429,7 @@ spec:
   minKubeVersion: 1.8.0
   provider:
     name: Red Hat
-  version: 0.0.0
+  version: 0.10.0
   webhookdefinitions:
     - admissionReviewVersions:
         - v1beta1
@@ -443,3 +443,4 @@ spec:
       targetPort: 9443
       type: ConversionWebhook
       webhookPath: /convert
+  replaces: authorino-operator.v0.9.0
diff --git a/bundle/metadata/annotations.yaml b/bundle/metadata/annotations.yaml
index 118a56d0..e8e53507 100644
--- a/bundle/metadata/annotations.yaml
+++ b/bundle/metadata/annotations.yaml
@@ -4,7 +4,8 @@ annotations:
   operators.operatorframework.io.bundle.manifests.v1: manifests/
   operators.operatorframework.io.bundle.metadata.v1: metadata/
   operators.operatorframework.io.bundle.package.v1: authorino-operator
-  operators.operatorframework.io.bundle.channels.v1: alpha
+  operators.operatorframework.io.bundle.channels.v1: stable
+  operators.operatorframework.io.bundle.channel.default.v1: stable
   operators.operatorframework.io.metrics.builder: operator-sdk-v1.22.0
   operators.operatorframework.io.metrics.mediatype.v1: metrics+v1
   operators.operatorframework.io.metrics.project_layout: go.kubebuilder.io/v3
diff --git a/config/authorino/kustomization.yaml b/config/authorino/kustomization.yaml
index ef0ddde0..31af44ab 100644
--- a/config/authorino/kustomization.yaml
+++ b/config/authorino/kustomization.yaml
@@ -2,14 +2,14 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
 resources:
-- github.com/Kuadrant/authorino/install?ref=main
+- github.com/Kuadrant/authorino/install?ref=v0.16.0
 - webhook
 
 # Configures the conversion webhook
 images:
 - name: AUTHORINO_IMAGE
   newName: quay.io/kuadrant/authorino
-  newTag: latest
+  newTag: v0.16.0
 
 patchesStrategicMerge:
 - webhook/patches/webhook_in_authconfigs.yaml
diff --git a/config/deploy/manifests.yaml b/config/deploy/manifests.yaml
index 9c262bba..a7f28948 100644
--- a/config/deploy/manifests.yaml
+++ b/config/deploy/manifests.yaml
@@ -6150,7 +6150,7 @@ spec:
         - --leader-elect
         command:
         - /manager
-        image: quay.io/kuadrant/authorino-operator:latest
+        image: quay.io/kuadrant/authorino-operator:v0.10.0
         livenessProbe:
           httpGet:
             path: /healthz
@@ -6201,7 +6201,7 @@ spec:
       - command:
         - authorino
         - webhooks
-        image: quay.io/kuadrant/authorino:latest
+        image: quay.io/kuadrant/authorino:v0.16.0
         name: webhooks
         ports:
         - containerPort: 9443
diff --git a/config/manifests/bases/authorino-operator.clusterserviceversion.yaml b/config/manifests/bases/authorino-operator.clusterserviceversion.yaml
index 595b71ad..455682df 100644
--- a/config/manifests/bases/authorino-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/authorino-operator.clusterserviceversion.yaml
@@ -5,13 +5,13 @@ metadata:
     alm-examples: '[]'
     capabilities: Basic Install
     categories: Integration & Delivery
-    containerImage: quay.io/kuadrant/authorino-operator:latest
+    containerImage: quay.io/kuadrant/authorino-operator:v0.10.0
     createdAt: 2021-12-08T10-00-00Z
     operators.operatorframework.io/builder: operator-sdk-v1.9.0
     operators.operatorframework.io/project_layout: unknown
     repository: https://github.com/Kuadrant/authorino-operator
     support: kuadrant
-  name: authorino-operator.v0.0.0
+  name: authorino-operator.v0.10.0
   namespace: placeholder
 spec:
   apiservicedefinitions: {}
@@ -67,4 +67,4 @@ spec:
   minKubeVersion: 1.8.0
   provider:
     name: Red Hat
-  version: 0.0.0
+  version: 0.10.0
```

Commit the changes as if you were simulating the creation of the release branch:

```sh
git checkout -b simulation-v0.10.0
git add -Av
git commit -m "Simulate release v0.10.0"
```

Verify the bundle manifests:

```sh
make verify-bundle VERSION=0.10.0 AUTHORINO_VERSION=0.16.0 CHANNELS=stable DEFAULT_CHANNEL=stable
```
```
git diff -I'^    containerImage:' -I'^                image:' --exit-code ./bundle
[ -z "$(git ls-files --other --exclude-standard --directory --no-empty-directory ./bundle)" ]
yq e -e '.metadata.annotations.containerImage == "quay.io/kuadrant/authorino-operator:v0.10.0"' bundle/manifests/authorino-operator.clusterserviceversion.yaml
true
yq e -e '.spec.install.spec.deployments[0].spec.template.spec.containers[0].image == "quay.io/kuadrant/authorino-operator:v0.10.0"' bundle/manifests/authorino-operator.clusterserviceversion.yaml
true
yq e -e '.spec.install.spec.deployments[1].spec.template.spec.containers[0].image == "quay.io/kuadrant/authorino:v0.16.0"' bundle/manifests/authorino-operator.clusterserviceversion.yaml
true
yq e -e '.spec.replaces == "authorino-operator.v0.9.0"' bundle/manifests/authorino-operator.clusterserviceversion.yaml
true
```

Makefile
guicassolato: About `.spec.replaces`, you're right. We still need it for the Authorino image though.

Makefile
adam-cattermole: Yep you're right we should just update the file indentation once and from then on it'll be clearer what's going on in each new release branch.

Makefile
guicassolato: You can test it with the [verification steps](https://github.com/Kuadrant/authorino-operator/pull/161#issuecomment-1825514169) I left. Instead of verifying the bundle with the correct `AUTHORINO_VERSION`, try a different one (only in the last step). E.g.:

```sh
make verify-bundle VERSION=0.10.0 AUTHORINO_VERSION=0.17.0 CHANNELS=stable DEFAULT_CHANNEL=stable
```

It will pass as if the bundle was valid while it should not.

Makefile
```diff
@@ -312,11 +326,11 @@ verify-manifests: manifests $(YQ) ## Verify manifests update.
 	yq e -e '.metadata.annotations.containerImage == "$(OPERATOR_IMAGE)"' config/manifests/bases/authorino-operator.clusterserviceversion.yaml
```
guicassolato, line 326: I think line 325 will fail since we have 2 deployments in the manifests now, i.e., `authorino-operator` and `authorino-webhooks`.

Suggestion to change it to:

```mk
	yq ea -e 'select([.][].kind == "Deployment") | select([.][].metadata.name == "authorino-operator").spec.template.spec.containers[0].image | . == "$(OPERATOR_IMAGE)"' config/deploy/manifests.yaml
	yq ea -e 'select([.][].kind == "Deployment") | select([.][].metadata.name == "authorino-webhooks").spec.template.spec.containers[0].image | . == "$(DEFAULT_AUTHORINO_IMAGE)"' config/deploy/manifests.yaml
```

guicassolato: I've updated the verification steps accordingly.

Makefile
```diff
@@ -312,11 +326,11 @@ verify-manifests: manifests $(YQ) ## Verify manifests update.
...
 
 .PHONY: verify-bundle
-verify-bundle: bundle $(YQ) ## Verify bundle update.
+verify-bundle: $(YQ) ## Verify bundle update.
 	git diff -I'^    containerImage:' -I'^                image:' --exit-code ./bundle
 	[ -z "$$(git ls-files --other --exclude-standard --directory --no-empty-directory ./bundle)" ]
-	yq e -e '.metadata.annotations.containerImage == "$(OPERATOR_IMAGE)"' bundle/manifests/authorino-operator.clusterserviceversion.yaml
-	yq e -e '.spec.install.spec.deployments[0].spec.template.spec.containers[0].image == "$(OPERATOR_IMAGE)"' bundle/manifests/authorino-operator.clusterserviceversion.yaml
+	yq e -e '.metadata.annotations.containerImage == "$(OPERATOR_IMAGE)"' $(BUNDLE_CSV)
+	yq e -e '.spec.install.spec.deployments[0].spec.template.spec.containers[0].image == "$(OPERATOR_IMAGE)"' $(BUNDLE_CSV)
```
guicassolato, line 333:
```suggestion
	$(YQ) e -e '.metadata.annotations.containerImage == "$(OPERATOR_IMAGE)"' $(BUNDLE_CSV)
	$(YQ) e -e '.spec.install.spec.deployments[0].spec.template.spec.containers[0].image == "$(OPERATOR_IMAGE)"' $(BUNDLE_CSV)
```

Makefile
```diff
@@ -83,8 +86,9 @@ AUTHORINO_BRANCH = v$(AUTHORINO_VERSION)
 AUTHORINO_IMAGE_TAG = v$(AUTHORINO_VERSION)
 endif
 
-AUTHORINO_IMAGE_FILE ?= authorino_image
-DEFAULT_AUTHORINO_IMAGE ?= $(shell cat $(AUTHORINO_IMAGE_FILE) || echo $(DEFAULT_REGISTRY)/$(DEFAULT_ORG)/authorino:latest)
+# Build file used to store replaces/authorinoImage options.
+BUILD_CONFIG_FILE ?= build.yaml
+DEFAULT_AUTHORINO_IMAGE ?= $(shell yq e -e '.config.authorinoImage' $(BUILD_CONFIG_FILE) || echo $(DEFAULT_REGISTRY)/$(DEFAULT_ORG)/authorino:latest)
```
guicassolato, line 91: I guess this will fail if you don't have `yq` in your `PATH` (even if the build config file exists. Maybe `$(YQ)` will do the job?

Makefile
adam-cattermole: Sure, but I think it'll pass anyway with the above change because `DEFAULT_AUTHORINO_IMAGE` is not affected by `AUTHORINO_VERSION` passed at command line, only whether the file is present or not.

guicassolato: Right. So maybe we can define:

```make
EXPECTED_DEFAULT_AUTHORINO_IMAGE = $(DEFAULT_REGISTRY)/$(DEFAULT_ORG)/authorino:$(AUTHORINO_IMAGE_TAG)
```

And then:

```make
verify-manifests:
	[…]
	$(YQ) ea -e 'select([.][].kind == "Deployment") | select([.][].metadata.name == "authorino-webhooks").spec.template.spec.containers[0].image | . == "$(EXPECTED_DEFAULT_AUTHORINO_IMAGE)"' config/deploy/manifests.yaml
```

and

```make
verify-bundle:
	[…]
	$(YQ) e -e '.spec.install.spec.deployments[1].spec.template.spec.containers[0].image == "$(EXPECTED_DEFAULT_AUTHORINO_IMAGE)"' bundle/manifests/authorino-operator.clusterserviceversion.yaml
```

WDYT?

adam-cattermole: I'm not sure I see the value in this tbh - happy to put it in if you like. If the bundle command is run prior and the file is present in the release branch the git diff will catch the change anyway. This additional check will ensure that if you pass the `AUTHORINO_VERSION` variable to `verify-` commands that doesn't match the one in the file that has been generated then you fail. This would only happen if for some reason the version in the file has been generated at a different version to what you're running the verify at right? is there a case I'm missing where this is likely to happen and be a problem?

Makefile
```diff
@@ -312,11 +326,11 @@ verify-manifests: manifests $(YQ) ## Verify manifests update.
 	yq e -e '.metadata.annotations.containerImage == "$(OPERATOR_IMAGE)"' config/manifests/bases/authorino-operator.clusterserviceversion.yaml
 
 .PHONY: verify-bundle
-verify-bundle: bundle $(YQ) ## Verify bundle update.
```
adam-cattermole, line 315: I think we can add this dependency back in again now we're using the build file

Makefile
adam-cattermole: Yep I set $(YQ) here but I think this means any run of the makefile is now a dependency on $(YQ)

Makefile
adam-cattermole: I've added this change anyway we can remove it later if it's unnecessary I guess

guicassolato: The way this goes is: first you generate the manifests (a set of params is given here), then you commit the changes, then you verify.

As for the verification itself, it goes: you generate the manifests again (supposedly with the exact same set of params used before), then you check if there's a diff. If there is a diff, then verification fails; otherwise, verification passes and the manifests are considered correct.

When checking the diff, however, it ignores all lines in the diff containing `    containerImage:` or `        image:`. I'm now struggling to see why we ignore these lines. Additionally, have we checked the indentation used here since we moved to using `yq`?

In the end, I think the purpose of verifying the manifests (both base and final bundle) is checking if the set of params used when the manifests were generated was correct.

adam-cattermole: > When checking the diff, however, it ignores all lines in the diff containing     containerImage: or         image:. I'm now struggling to see why we ignore these lines. Additionally, have we checked the indentation used here since we moved to using yq?

IIRC this is to overcome the 'chicken and egg' issue of creating builds with the containerImages set to the commit sha build of the operator. We only know the commit sha after it has been committed, so the CSV cannot contain the commit sha. Instead we check that the `OPERATOR_IMAGE` containing the commit sha is in the CSV.

Good point on the whitespace though, it looks like it needs updating will do that now.

Makefile
```diff
@@ -255,22 +263,29 @@ bundle-build: ## Build the bundle image.
...
-.PHONY: fix-csv-replaces
-fix-csv-replaces: $(YQ)
+.PHONY: create-build-file
+create-build-file: $(YQ)
+	$(YQ) -n '.config' > $(BUILD_CONFIG_FILE)
+
+.PHONY: set-authorino-default-image
+set-authorino-default-image: $(YQ)
+	@if [ "$(AUTHORINO_VERSION)" != "latest" ]; then\
+		V="$(DEFAULT_REGISTRY)/$(DEFAULT_ORG)/authorino:$(AUTHORINO_IMAGE_TAG)" $(YQ) eval '.config.authorinoImage = strenv(V)' -i $(BUILD_CONFIG_FILE); \
```
guicassolato, line 273: I see `.github/workflows/build-images.yaml:57` still refers to the old `authorino_image` file.

Makefile
```diff
@@ -308,15 +323,17 @@ catalog-push: ## Push a catalog image.
 verify-manifests: manifests $(YQ) ## Verify manifests update.
 	git diff -I'^    containerImage:' -I'^        image:' --exit-code ./config
```
guicassolato, line 324:
```suggestion
	git diff -I' containerImage:' -I' image:' --exit-code ./config
```

Makefile
```diff
@@ -308,15 +323,17 @@ catalog-push: ## Push a catalog image.
...
-	yq ea -e 'select([.][].kind == "Deployment").spec.template.spec.containers[0].image | . == "$(OPERATOR_IMAGE)"' config/deploy/manifests.yaml
-	yq e -e '.metadata.annotations.containerImage == "$(OPERATOR_IMAGE)"' config/manifests/bases/authorino-operator.clusterserviceversion.yaml
+	$(YQ) ea -e 'select([.][].kind == "Deployment") | select([.][].metadata.name == "authorino-operator").spec.template.spec.containers[0].image | . == "$(OPERATOR_IMAGE)"' config/deploy/manifests.yaml
+	$(YQ) ea -e 'select([.][].kind == "Deployment") | select([.][].metadata.name == "authorino-webhooks").spec.template.spec.containers[0].image | . == "$(EXPECTED_DEFAULT_AUTHORINO_IMAGE)"' config/deploy/manifests.yaml
+	$(YQ) e -e '.metadata.annotations.containerImage == "$(OPERATOR_IMAGE)"' config/manifests/bases/authorino-operator.clusterserviceversion.yaml
 
 .PHONY: verify-bundle
 verify-bundle: bundle $(YQ) ## Verify bundle update.
-	git diff -I'^    containerImage:' -I'^                image:' --exit-code ./bundle
+	git diff -I'^    containerImage:' -I'^                    image:' --exit-code ./bundle
```
guicassolato, line 332:
```suggestion
	git diff -I' containerImage:' -I' image:' --exit-code ./bundle
```

adam-cattermole: We can do it like this, but we'll start ignoring additional lines - for example the images in the scorecard folder would now match as well.

guicassolato: No need. I see you've already fixed it.

guicassolato (approved):
Excellent job, @adam-cattermole! Thank you!

PR #163, opened 2023-12-05
Add missing information to CSV

### LINKED ISSUE #162: Missing AuthConfig v1beta2 display name and description in the CSV
When the OLM manifest bundle is [generated](https://github.com/Kuadrant/authorino-operator/blob/a3fa8226e216bf7157e81790a65b251372ca5a98/Makefile#L250), for some reason the AuthConfig v1beta2 entry in the CSV misses display name and description:

https://github.com/Kuadrant/authorino-operator/blob/a3fa8226e216bf7157e81790a65b251372ca5a98/bundle/manifests/authorino-operator.clusterserviceversion.yaml#L46-L48

This is causing the operator to show up in the operator hub like this:

Screenshot 2023-11-28 at 18 42 03

However, those values seem to be properly set in the source code ([v1beta1](https://github.com/Kuadrant/authorino/blob/9da0bf6d9afce55ec3ff280fe7c28b54e7a5e6ab/api/v1beta1/auth_config_types.go#L784), [v1beta2](https://github.com/Kuadrant/authorino/blob/9da0bf6d9afce55ec3ff280fe7c28b54e7a5e6ab/api/v1beta2/auth_config_types.go#L91)) and generated CRD specs ([v1beta1](https://github.com/Kuadrant/authorino/blob/9da0bf6d9afce55ec3ff280fe7c28b54e7a5e6ab/install/crd/authorino.kuadrant.io_authconfigs.yaml#L55), [v1beta2](https://github.com/Kuadrant/authorino/blob/9da0bf6d9afce55ec3ff280fe7c28b54e7a5e6ab/install/crd/authorino.kuadrant.io_authconfigs.yaml#L2414C4-L2414C4).) Therefore, it could be a problem of when generating the bundle only. TBC.

grzpiotrowski:
This PR attempts to fix #162.

It looks like adding the information to the CSV template sorts out the issue of missing names/descriptions.

There is one peculiar thing caused by two versions of AuthConfig CRD which results in two seemingly duplicated cards rendered in OperatorHub.

While the displayed name could be altered so the first card would display for example `AuthConfig/v1beta1` and the second card `AuthConfig/v1beta2` the YAML example displayed on both cards would still show an example from `AuthConfig/v1beta2` (even for v1beta1 card), despite examples existing for both versions in the CSV file. It's a bit confusing, I hope this describes the issue in an understandable way.

For this reason I have included the example YAML for `AuthConfig/v1beta2` only and kept the `displayName` for both versions as `AuthConfig`. This way the OperatorHub will only show the most recent version to the visitors with a correct YAML example.

As an example, this approach can be also seen in [ArgoCD operator](https://operatorhub.io/operator/argocd-operator) with the `Argo CD` schema.

To verify the changes, copy the full contents of the generated CSV manifest to [OperatorHub Preview](https://operatorhub.io/preview) tool.

Also a pic of how it looks like after the changes:
image

### DIFF
--- a/config/manifests/bases/authorino-operator.clusterserviceversion.template.yaml
+++ b/config/manifests/bases/authorino-operator.clusterserviceversion.template.yaml
@@ -21,6 +21,11 @@ spec:
       displayName: AuthConfig
       kind: AuthConfig
       name: authconfigs.authorino.kuadrant.io
+      version: v1beta2
+    - description: API to describe the desired protection for a service
+      displayName: AuthConfig
+      kind: AuthConfig
+      name: authconfigs.authorino.kuadrant.io
       version: v1beta1
     - description: API to create instances of authorino
       displayName: Authorino
--- a/bundle/manifests/authorino-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/authorino-operator.clusterserviceversion.yaml
@@ -5,6 +5,34 @@ metadata:
     alm-examples: |-
       [
         {
+          "apiVersion": "authorino.kuadrant.io/v1beta2",
+          "kind": "AuthConfig",
+          "metadata": {
+            "name": "talker-api-protection"
+          },
+          "spec": {
+            "authentication": {
+              "api-key-users": {
+                "apiKey": {
+                  "selector": {
+                    "matchLabels": {
+                      "group": "friends"
+                    }
+                  }
+                },
+                "credentials": {
+                  "authorizationHeader": {
+                    "prefix": "APIKEY"
+                  }
+                }
+              }
+            },
+            "hosts": [
+              "talker-api.io"
+            ]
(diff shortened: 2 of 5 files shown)


config/manifests/bases/authorino-operator.clusterserviceversion.template.yaml
```diff
@@ -17,6 +17,11 @@ spec:
   apiservicedefinitions: {}
   customresourcedefinitions:
     owned:
+    - description: API to describe the desired protection for a service
```
eguzki, line 20: What is this template?? where is it being used? How?

eguzki (approved):

guicassolato (approved):
Nice! Thanks, @grzpiotrowski!

PR #164, opened 2023-12-08
Add issues/PRs to new Project

alexsnaps:


.github/workflows/issues-workflow.yaml
```diff
@@ -4,6 +4,9 @@ on:
   issues:
     types:
       - opened
+  pull_request:
```
guicassolato, line 7: Same as https://github.com/Kuadrant/authorino/pull/449/files#r1422236195

alexsnaps: Same as https://github.com/Kuadrant/authorino/pull/449/files#r1422413241 🤣

guicassolato (approved):

PR #165, opened 2024-01-12
Adding MAINTAINERS/ADOPTERS files

guicassolato:


ADOPTERS
KevFan, line 1: nit: Should these be using `.md` extension like in https://github.com/Kuadrant/authorino ? 🤔

guicassolato: Nice catch! Thanks.

didierofrivia (approved):
👯

PR #170, opened 2024-02-13
Add example for AuthConfig v1beta1 in the CSV

guicassolato:
What: Adds an example of the AuthConfig CR in v1beta1 to the CSV
Why: Having an example for each version of the API that is served is required by some operator hub test pipelines.


didierofrivia (approved):
🍸

PR #184, opened 2024-06-24
Helm charts

didierofrivia:
This PR introduces a way to manage an Authorino Operator [Helm Chart](https://helm.sh/). This is not meant to replace the way we are building and delivering our manifests (Kustomize, OLM) but to provide an alternative (complementary) way of delivering the operator.

This early implementation uses Kustomize to create the chart template, instead of creating and maintaining new ones with Helm, to later customize the Helm only settings with its `values.yaml`

**NOTES**

* This chart (and every other operator chart we package in the future: Limitador, DNS, etc) bundles only the operator in the same fashion OLM does.
* The operator version matches the chart version
* This first release doesn't use helm templating, it provides only the manifests previously built by Kustomize.
* The sole repository for all our operator charts will be hosted in [Kuadrant/helm-charts](https://github.com/Kuadrant/helm-charts/).
* The "source of truth" (source code and releasing logic) will be provided in each operator repo, not in the helm charts repository above.
* The helm charts repository is [Kuadrant/helm-charts](https://github.com/Kuadrant/helm-charts/)
* Documentation will be provided once the repository and other operator charts are tested.
* The steps to try out the repository will be provided in the kuadrant helm charts repo.
* For the release of the package, it will be tested once merged with alpha pre releases.

**Verification Steps**

0. Install helm locally
```sh
make helm
```

1. Create a local kind cluster
```sh
kind create cluster --name kuadrant-local
```

2. Install required dependency cert-manager
```sh
kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.12.1/cert-manager.crds.yaml
```

```sh
# Add the Jetstack Helm repository
bin/helm repo add jetstack https://charts.jetstack.io --force-update

# Install the cert-manager helm chart
bin/helm install cert-manager --namespace cert-manager --version v1.12.1 jetstack/cert-manager --create-namespace
```

You should see it installed:
```sh
kubectl get deployments -n cert-manager
NAME                      READY   UP-TO-DATE   AVAILABLE   AGE
cert-manager              1/1     1            1           7m32s
cert-manager-cainjector   1/1     1            1           7m32s
cert-manager-webhook      1/1     1            1           7m32s
```

3. Deploy (install) the default (`latest`) Authorino Operator chart
```sh
make helm-install
```

OR

```sh
bin/helm install authorino-operator charts/authorino-operator
```

4. Verify the installed operator image:
```sh
kubectl get deployments/authorino-operator -n authorino-operator -o yaml | grep image:
```
it should return:
```sh
image: quay.io/kuadrant/authorino-operator:latest
```

It should also have installed Authorino manifests and webhooks i.e:

```sh
kubectl get crds | grep auth
authconfigs.authorino.kuadrant.io           2024-07-01T08:26:05Z
authorinos.operator.authorino.kuadrant.io   2024-07-01T08:26:05Z
```

```sh
kubectl get deployment -n authorino-operator

NAME                 READY   UP-TO-DATE   AVAILABLE   AGE
authorino-operator   1/1     1            1           2m6s
authorino-webhooks   1/1     1            1           2m6s
```

5. Build new manifests with specific version of the operator
```sh
make helm-build VERSION=0.12.0-alpha1 AUTHORINO_VERSION=0.17.2
```

6. Upgrade the current installed operator to the freshly built one
```sh
make helm-upgrade
```

7. Check the installed operator image:
```sh
kubectl get deployments/authorino-operator -n authorino-operator -o yaml | grep image:
```

it should return:
```sh
image: quay.io/kuadrant/authorino-operator:v0.12.0-alpha1
```

8. For uninstalling the chart:
```sh
make helm-uninstall
```


guicassolato (changes requested):
Installing Authorino Operator should also install the base resources that allows deploying Authorino instances later on, including a Deployment of authorino-webhooks, as well as registering the AuthConfig CRD and a couple of ClusterRoles.

Because of the webhooks, there's also a dependency on [cert-manager](https://cert-manager.io/) (when not installing via OLM.)

The manifests of reference for all the above are: https://github.com/Kuadrant/authorino-operator/blob/main/config/deploy/manifests.yaml, along with the install script: https://github.com/Kuadrant/authorino-operator/blob/main/utils/install.sh.

I imagine that installing Authorino Operator via Helm chart should be a replacement to the install script.

On a different take... If we're planning on having an "Authorino Helm chart" too, then the Authorino one needs to be introduced first, and then this one with a dependency on it. The Authorino Helm chart could then install the missing bits of this PR, without which installing Authorino Operator would only result in broken deployments of the operand.

Makefile
```diff
@@ -139,6 +139,23 @@ OPM = $(shell which opm)
...
 endif
 
+HELM = ./bin/helm
+HELM_VERSION = v3.15.0
+$(HELM):
+	@{ \
+	set -e ;\
+	mkdir -p $(dir $(HELM)) ;\
+	OS=$(shell go env GOOS) && ARCH=$(shell go env GOARCH) && \
+	wget -O helm.tar.gz https://get.helm.sh/helm-$(HELM_VERSION)-$${OS}-$${ARCH}.tar.gz ;\
```
guicassolato, line 149: Verification steps failed to me due to missing `wget`. Can we use curl instead, since there are other commands that depend on it too?

```
$ make helm-install
bash: wget: command not found
make: *** [bin/helm] Error 127
```

guicassolato: On a separate question, can we rename `charts/crds` to something like `charts/dependencies` or so?

cert-manager's manifests include a lot more than CRDs (namespaces, deployments, etc.) Additionally, authorino-operator's manifests (the one under `charts/templates`) also include CRDs, yet it's not in `charts/crds`.

make/helm.mk
```diff
@@ -0,0 +1,60 @@
...
+
+.PHONY: helm-build
+helm-build: $(YQ) kustomize manifests ## Build the helm chart from kustomize manifests
+	# Replace the controller image
+	cd config/manager && $(KUSTOMIZE) edit set image controller=$(OPERATOR_IMAGE)
+	# Build the helm chart templates from kustomize manifests
+	$(KUSTOMIZE) build config/helm > charts/authorino-operator/templates/manifests.yaml
+	V="$(BUNDLE_VERSION)" $(YQ) -i e '.version = strenv(V)' charts/authorino-operator/Chart.yaml
+	V="$(CERT_MANAGER_VERSION)" $(YQ) -i e '(.dependencies[] | select(.name == "cert-manager").version) = strenv(V)' -i charts/authorino-operator/Chart.yaml
+	wget -O charts/authorino-operator/crds/cert-manager-manifests.yaml https://github.com/cert-manager/cert-manager/releases/download/v${CERT_MANAGER_VERSION}/cert-manager.yaml
```
guicassolato, line 11: Can we rename `charts/crds` to something like `charts/dependencies` or so?

cert-manager's manifests include a lot more than CRDs (namespaces, deployments, etc.) Additionally, authorino-operator's manifests (the one under `charts/templates`) also include CRDs, yet it's not in `charts/crds`.

Makefile
guicassolato:
```suggestion
	curl -sL -o helm.tar.gz https://get.helm.sh/helm-$(HELM_VERSION)-$${OS}-$${ARCH}.tar.gz ;\
```

guicassolato:
Verification steps seem to be applying 2 sets of deployments of cert-manager. One set in the `cert-manager` namespace and another in the `default` namespace. The deployments in the `default` namespace are prefixed by `authorino-operator-`. This seems redundant. Normally, we'd only have it deployed to the `cert-manager`.

make/helm.mk
didierofrivia: It was my fault regarding the set of manifests installed related to cert-manager, now it's installed just the CRDs. 
Regarding the `charts/crds` directory, Helm is using that directory to install any needed CRD before its mechanism of installing charts and its dependencies kicks in. I thought it was a good place to host the cert-manager CRDs that are require by its chart https://artifacthub.io/packages/helm/cert-manager/cert-manager

didierofrivia: Ok, decided to remove the dependency and add the extra steps, will be documented on the repo as well.

guicassolato (approved):
Verification steps work. Let's merge the PR. Thanks for the work, @didierofrivia!

As a side note, it's a shame that Helm decided to keep the installation of the CRDs as a separate step though. I can see how users could easily end up with mismatching versions between API and backend.

For now, I wouldn't replace the install script to default to `helm install cert-manager; helm install authorino-operator`, as I think the script providers users with better experience.

Maybe in the future there will be a better way to handle dependencies in Helm that doesn't involve deploying dedicated instances of cert-manager for Authorino Operator nor duplicate cluster-scope installs.

PR #186, opened 2024-06-28
Fix: Adding missing operator metrics service

R-Lawton:
Adding missing service to expose authorino operator metrics & adding flag to fix panic that happens with manifests make target

## Verification
```
kind create cluster --name authorino-operator
make install
make run
```

Create a authorino cr
```
kubectl apply -f -<<EOF             
kind: Authorino
metadata:
  name: authorino
  namespace: authorino-operator
spec:
  clusterWide: true
  healthz: {}
  listener:
    ports: {}
    tls:
      enabled: false
  metrics: {}
  oidcServer:
    tls:
      enabled: false
  supersedingHostSubsets: true
  tracing:
    endpoint: ''
  volumes: {}
EOF
```

You should see multiple services come up
`k get services -n authorino-operator
`
ensure you get the following
```
authorino-metrics                                      ClusterIP   10.96.96.154   <none>        8080/TCP             8s
authorino-authorino-authorization                       ClusterIP   10.96.218.59   <none>        50051/TCP,5001/TCP   8s
authorino-authorino-oidc                                ClusterIP   10.96.25.6     <none>        8083/TCP             8s
authorino-operator-controller-manager-metrics-service   ClusterIP   10.96.197.85   <none>        8080/TCP             8s
authorino-webhooks                                      ClusterIP   10.96.168.70   <none>        443/TCP              113s
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -142,7 +142,7 @@ endif
 ##@ Development
 
 manifests: controller-gen kustomize authorino-manifests ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
-	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=authorino-operator-manager webhook paths="./..." output:crd:artifacts:config=config/crd/bases && $(KUSTOMIZE) build config/install > $(OPERATOR_MANIFESTS)
+	$(CONTROLLER_GEN) -h $(CRD_OPTIONS) rbac:roleName=authorino-operator-manager webhook paths="./..." output:crd:artifacts:config=config/crd/bases && $(KUSTOMIZE) build config/install > $(OPERATOR_MANIFESTS)
 	$(MAKE) deploy-manifest OPERATOR_IMAGE=$(OPERATOR_IMAGE)
 
 .PHONY: authorino-manifests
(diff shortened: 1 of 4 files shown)


guicassolato (changes requested):
Thanks for the PR, @R-Lawton.

I think what we want is something along the lines this what we have already for Limitador Operator:
- https://github.com/Kuadrant/limitador-operator/blob/main/config/manager/metrics_service.yaml
- https://github.com/Kuadrant/limitador-operator/blob/main/config/default/manager_metrics_patch.yaml

No changes needed in the `AuthorinoReconciler` IMO. Just a simple 1-off config for when installing the operator itself.

guicassolato:
Just pushed #187 as an alternative.

<br/>

In time – one acknowledgement though to the bad suffix `controller-metrics` in the operands' metric services' names that this PR would help fix:

https://github.com/Kuadrant/authorino-operator/blob/d6405733559a44c8673b0c72f265052cd51bfec6/pkg/resources/k8s_services.go#L36

I think this bit is very welcome, though probably in a separate PR since it's a breaking change for the user.

R-Lawton:
Ok cool, i wasnt sure best way of going about it so i went with how the rest of the services are created but i think keeping it cohesive with limitador makes sense so ill close this out in favour of your pr

PR #187, opened 2024-06-28
Add Authorino Operator metrics service

guicassolato:


bundle/manifests/authorino-operator-metrics_v1_service.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+kind: Service
+metadata:
+  creationTimestamp: null
+  labels:
+    control-plane: authorino-operator
+  name: authorino-operator-metrics
```
R-Lawton, line 7: Im thinking we make the naming the same as it it for  limitador, kuadrant and cert manager to. `<OPERATOR_NAME>-controller-manager-metrics-service` wdyt?

R-Lawton:
Do you want to also add the change i had in the other pr with the makefile stopping controller-gen SIGSEGV panic

bundle/manifests/authorino-operator-metrics_v1_service.yaml
guicassolato: I agree with the premise, though IMHO I believe it's Limitador Operator the one that should change.

`controller-manager` is very poor choice for something that is known by the user as "\<X\> Operator", while the `-service` suffix is not needed to identity an object whose kind is already a Service.

guicassolato:
> Do you want to also add the change i had in the other pr with the makefile stopping controller-gen SIGSEGV panic

Is that en environment-specific issue? I could run `make manifests bundle` just fine locally.

R-Lawton:
No it wasnt just my machine I got Jason to run the same commands and he ran into the issue to

guicassolato:
> No it wasnt just my machine I got Jason to run the same commands and he ran into the issue to

I'm so sorry, @R-Lawton. You are absolutely right!

After `rm -rf ./bin/*`, I also see the SIGSEGV. I guess my controller-gen was outdated. Pushing your fix now...

bundle/manifests/authorino-operator-metrics_v1_service.yaml
R-Lawton: im neither here nor there about it, i think should all just be the same. I can change the other 3 services to follow suit of this naming if you prefer

config/default/manager_metrics_patch.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: apps/v1
+kind: Deployment
```
R-Lawton, line 2: Why are we patching the deployment with the port vs not adding it into the deployment when its being created?

bundle/manifests/authorino-operator-metrics_v1_service.yaml
guicassolato: Operator SDK's the one blame, I think. But yeah. I do prefer plain simple `authorino-operator-metrics` here. In fact, we went that road 2 years ago: #52.

config/default/manager_metrics_patch.yaml
guicassolato: TL;DR – Operator vs operand. It's a different Deployment.

The Operator's Deployment is created just once, when the Operator is installed. This is the Deployment whose missing port number this PR fixes. This Deployment won't ever be affected by reconciliation. It has nothing to do with Authorino CRs.

That said, we _also_ do not specify the port numbers in the Deployments (plural) that the Operator creates for the operand[^1][^2].

[^1]: https://github.com/Kuadrant/authorino-operator/blob/1acbdfc2040210a85a2e2cb156a22add964c1414/controllers/authorino_controller.go#L285

[^2]: https://github.com/Kuadrant/authorino-operator/blob/1acbdfc2040210a85a2e2cb156a22add964c1414/pkg/resources/k8s_deployment.go#L38-L44

The problems are unrelated though, even if they are perceived the same way by one who only wants to consume the metrics from all components.

We can fix the lack of port numbers in the operands' Deployments too (in another PR.)

Either way, let's keep in mind that changing the Deployments is not what will affect exposure of the port. From the docs[^3]:

[^3]: https://pkg.go.dev/k8s.io/api@v0.28.3/core/v1#Container

>  Not specifying a port here DOES NOT prevent that port from being exposed. Any port which is listening on the default "0.0.0.0" address inside a container will be accessible from the network.

Creating the Services is the main piece in our case. We currently create (and reconcile!) the metrics services for the operand all right. With this PR, we will be also creating (though no reconciliation needed) the metrics service for the Operator.

Makefile
```diff
@@ -142,7 +142,7 @@ endif
 ##@ Development
 
 manifests: controller-gen kustomize authorino-manifests ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
-	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=authorino-operator-manager webhook paths="./..." output:crd:artifacts:config=config/crd/bases && $(KUSTOMIZE) build config/install > $(OPERATOR_MANIFESTS)
+	$(CONTROLLER_GEN) -h $(CRD_OPTIONS) rbac:roleName=authorino-operator-manager webhook paths="./..." output:crd:artifacts:config=config/crd/bases && $(KUSTOMIZE) build config/install > $(OPERATOR_MANIFESTS)
```
guicassolato, line 145: This feels weird. It's printing controller-gen's help for everything now.

Is it really how we want to fix the SIGSEGV issue?

config/default/manager_metrics_patch.yaml
guicassolato: Hahaha. Now I get what you mean!

> why not have the port in the deployment in the manager.yaml.

I suppose we could. Here I was the one following Limitador Operator's approach 🙃

Makefile
R-Lawton: I tried updating to new versions of controller gen and it fixes the panic issue but a new issue crops up that work with the -h. My thinking was go with -h for now to unblock quickly  and then try find a solution for other issue i think ill have to create a issue with them. My hunch is a versioning issue between go version and controller tools versions.  The panic is a known issue https://github.com/kubernetes-sigs/controller-tools/issues/888

config/default/manager_metrics_patch.yaml
R-Lawton: I feel like its an unnecessary step, i don't mind updating limitador to follow suit with authorino and update it as part of the service pr i, creating for limitador.

guicassolato: Done.

Makefile
guicassolato: I've upgraded controller-gen to v0.15.0 and apparently it worked without hiccup.

Do you mind to give #188 a go please? Make sure to first `rm -f ./bin/*`. If it works for you, I'd rather merge that one first, then clean this one up a bit.

guicassolato: Fixed in #188 and removed from this PR.

R-Lawton: Deadly happy days

R-Lawton (approved):
Looks great

PR #188, opened 2024-06-28
Upgrade controller-gen to v0.15.0

guicassolato:
Upgrades controller-gen to v0.15.0.

Among other things, fixes a SIGSEGV with previous v0.6.1 version beign used.


adam-cattermole (approved):
lgtm

Issue #3, opened 2021-11-08
Add a command to the makefile remove Authorino CRD


jjaferson:
Add a make command to the makefile for removing Authorino's CRDs, an example on how to remove the CRDs can be found [here](https://github.com/Kuadrant/authorino/blob/main/Makefile#L70)

_Originally posted by @guicassolato in https://github.com/Kuadrant/authorino-operator/pull/1#discussion_r741955437_

PR #191, opened 2024-07-08
Update make uninstall 

Boomatang:
closes: https://github.com/Kuadrant/authorino-operator/issues/3

- create and delete authorino-operator namespace (requied by make install-authorino)
- add make uninstall-authorino
- add make uninstall-cert-manager
- update make uninstall

# Verification 
- create a kind cluster
- list CRDS on cluster `kubectl get crds`
- install the authorino-operator resources `make install`
- check the CRDS again `kubectl get crds`
- run uninstall command `make uninstall`
- check the CRDS again `kubectl get crds`, this will equal the initial list of CRDS from the cluster.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -206,12 +206,9 @@ docker-push: ## Push docker image with the manager.
 
 ##@ Deployment
 
-install: manifests kustomize install-authorino ## Install CRDs into the K8s cluster specified in ~/.kube/config.
-	kubectl apply -f $(OPERATOR_MANIFESTS)
+install: install-authorino install-operator ## Install CRDs into the K8s cluster specified in ~/.kube/config.
 
-uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified in ~/.kube/config.
-	kubectl delete -f $(OPERATOR_MANIFESTS)
-	$(MAKE) uninstall-authorino
+uninstall: uninstall-operator uninstall-authorino ## Uninstall CRDs from the K8s cluster specified in ~/.kube/config.
 
 deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${OPERATOR_IMAGE}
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -211,6 +211,7 @@ install: manifests kustomize install-authorino ## Install CRDs into the K8s clus
 
 uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified in ~/.kube/config.
 	kubectl delete -f $(OPERATOR_MANIFESTS)
+	$(MAKE) uninstall-authorino
```
guicassolato, line 214: `make install` does not install Authorino CRDs. I wonder if `make uninstall` should indeed try to undo what its counterpart did not do.

Makefile
```diff
@@ -222,12 +223,21 @@ undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/confi
 	$(KUSTOMIZE) build config/default | kubectl delete -f -
 
 install-authorino: install-cert-manager ## install RBAC and CRD for authorino
+	kubectl create namespace authorino-operator
 	$(KUSTOMIZE) build config/authorino | kubectl apply -f -
 
+uninstall-authorino:  ## uninstall RBAC and CRD for authorino
+	$(KUSTOMIZE) build config/authorino | kubectl delete -f -
+	kubectl delete namespace authorino-operator
```
guicassolato, line 231: Deleting the operator's namespace seems a bit harsh. Effectively, this will decommission the operator instance. IWO, it's almost like a "undeploy" target.

Makefile
Boomatang: The `make install` does install the Authorino CRD, it is part of `OPERATOR_MANIFESTS`. The adding of the `make uninstall-authorino` is to mirror how the `make install` calls `install-authorino`. This change aligns closer with what its counterpart does.

Makefile
Boomatang: Ya the namespace thing was something that kept biting me when doing this. As the create of the namespace is documented in the README I would be happy the remove the changes I made around the namespace.

Makefile
```diff
@@ -222,12 +223,21 @@ undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/confi
...
+	$(MAKE) uninstall-cert-manager
```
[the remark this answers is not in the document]
Boomatang: The `make install-authorino` does call `install-cert-manager`. This is aligning the install and uninstall targets.

Makefile
guicassolato: My confusion. [`OPERATOR_MANIFESTS`](https://github.com/Kuadrant/authorino-operator/blob/a8652389ae4de98ca580198594f5a3198080b6de/Makefile#L53) points to [this](https://github.com/Kuadrant/authorino-operator/blob/a8652389ae4de98ca580198594f5a3198080b6de/config/install/manifests.yaml) file, which does _not_ include Authorino CRDs. However, `install` _depends on_ `install-authorino`. Makes sense.

Makefile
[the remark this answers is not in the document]
guicassolato: I see that now.

Tricky one. Installing when it's already present causes less harm than uninstalling when you aren't supposed to. Not sure what best here.

Makefile
Boomatang: I feel like I am missing something. The very first file in that manifest is the [authorino CRD](https://github.com/Kuadrant/authorino-operator/blob/a8652389ae4de98ca580198594f5a3198080b6de/config/install/manifests.yaml#L2-L13). What am I missing?

guicassolato: Authorino CRD belongs to the operator. AuthConfig CRD is the one that belongs to Authorino.

Makefile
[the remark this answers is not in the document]
Boomatang: Hmm, I see your point. The question I find my self asking is, what is the purpose of the make targets? Are they how we recommend the user installs the operator in production? If so how do we handle upgrades? Or is the purpose to aid developers in the day to day develop of the operator? In which case it would be better to revert all changes to a cluster? I am also not sure which is best but I do lean to the latter.

Makefile
Boomatang: Okay I get you now when you said Authorio CRDs you meant Authorino as in Authorino the product and not the CRD kind. Happy that confusion was cleared up. 
So to be sure, are you asking for anything to be changed here?

Makefile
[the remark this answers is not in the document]
guicassolato: > what is the purpose of the make targets?

Mainly dev and testing workflows, I'd say.

> Are they how we recommend the user installs the operator in production? 

No.

> Or is the purpose to aid developers in the day to day develop of the operator?

Yes.

> it would be better to revert all changes to a cluster? I am also not sure which is best but I do lean to the latter.

I was thinking... These make targets are useful for people who clone the repo, maybe they make changes, and want to try them out locally. They may no know exactly which YAMLs to apply for each step to of the workflow and therefore the targets provide them with a bit of guidance/documentation.

We cannot predict the exact scenario that development/test work is going on. E.g. maybe it's for a change in the controller's code, maybe it's a change in the Authorino CRD, maybe it's because of a change in Authorino itself that requires some tweak in the operator to be tested, maybe it's running in a fresh newly created cluster, maybe it's running on a preexisting cluster with preexisting stuff installed, maybe cert-manager is already running (installed for something else), maybe it's the first time the operator is installed, maybe it's a re-install, maybe it's to test a new version of CRDs on a preexisting install, etc.

With that in mind, perhaps a reasonable approach is providing very granular targets, for very specific operations, alongside with fewer more opinionated ones. For example,
- `install`: Installs Operator CRD/RBAC, plus Authorino CRD/RBAC
- `install-operator`: Installs Operator CRD/RBAC
- `install-authorino`: Installs Authorino CRD/RBAC
- `install-cert-manager`: Installs cert-manager
- `uninstall`: Uninstalls Operator CRD/RBAC, plus Authorino CRD/RBAC
- `uninstall-operator`: Uninstalls Operator CRD/RBAC
- `uninstall-authorino`: Uninstalls Authorino CRD/RBAC
- `uninstall-cert-manager`: Uninstalls cert-manager

- `namespace`: Creates `authorino-operator` namespace
- `delete-namespace`: Deletes `authorino-operator` namespace

- `deploy`: Creates Deployments of the Operator and webhooks; _depends on_ `install`, `cert-manager`, `namespace`
- `deploy-webhooks`: Creates Deployment of the webhook; _depends on_ `install-authorino`, `cert-manager`, `namespace`
- `undeploy`: Deletes Deployments of the Operator and webhooks
- `undeploy-webhooks`: Deletes Deployment of the webhook

- `run`: Runs the operator locally (with `go run`); _depends on_ `install`

I guess the pattern being:
- `<action>` = runs deps + action itself; safe to run if deps already present
- `un<action>` = undo the action only

WDYT?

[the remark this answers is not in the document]
Boomatang: Maybe, it is this idea of easy of use. The idea that `<action>` installs the required dependencies makes for a very easy of use. But not having the same easy of use for the `un<action>` seems wrong or at least creates a double standard. At the same time that issue of undoing something that was existing pre `<action>` is far worse. 

While I think this is a compromise and contradiction in usability, it is also possible the correct solution for now. Change would need some docs update also and the scope of some current targets would need to be reduced. The example off the top of my head is how the `install` target in a nest way also calls `install-cert-manager`. Which I think is not what your expecting to happen.

guicassolato (approved):
LGTM. Thanks, @Boomatang!

PR #193, opened 2024-07-22
Fix Helm workflow

didierofrivia:
Fixing typo due copy/pasta


guicassolato (approved):
Oops. I guess we all missed that. Well spotted, @didierofrivia. Thanks!

PR #195, opened 2024-08-15
[fix] simplifies test assertions

bartoszmajsak:
This change reworks functions called by Gomega's `Eventually` polling wrapper to return actual errors instead of return boolean expressions.

With this approach, error messages are more explicit, pointing to the exact error that unexpectedly occurred.

> [!NOTE]
> I also removed dead func


guicassolato (approved):
Thank you, @bartoszmajsak!

PR #198, opened 2024-08-21
Fixing bundle creation and verification

didierofrivia:
This PR aims to fix the bundle creation and verification when the tag is SEMVER. Assigning `VERSION` from the [GITHUB_REF_NAME](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables):

> The short ref name of the branch or tag that triggered the workflow run. This value matches the branch or tag name shown on GitHub. For example, feature-branch-1.
> ...

Removing the leading `v` in order to pass the expected version to the Makefile target.

Currently it was passing:
Screenshot 2024-08-21 at 18 15 55

Failing the job and not pushing the images.

### DIFF
--- a/.github/workflows/build-images.yaml
+++ b/.github/workflows/build-images.yaml
@@ -109,8 +109,9 @@ jobs:
       - name: Set Operator version
         id: operator-version
         run: |
-          if [[ ${GITHUB_REF_NAME/\//-} =~ ^v[0-9]+\.[0-9]+\.[0-9]+(-.+)?$ ]]; then
-          echo "VERSION=${GITHUB_REF_NAME/\//-}" >> $GITHUB_ENV
+          tag=${{GITHUB_REF_NAME}}
+          if [[ ${tag} =~ ^v[0-9]+\.[0-9]+\.[0-9]+(-.+)?$ ]]; then
+          echo "VERSION=${tag#v}" >> $GITHUB_ENV
           else
           echo "VERSION=${{ github.sha }}" >> $GITHUB_ENV
           fi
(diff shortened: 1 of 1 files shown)


.github/workflows/build-images.yaml
```diff
@@ -109,8 +109,9 @@ jobs:
       - name: Set Operator version
         id: operator-version
         run: |
-          if [[ ${GITHUB_REF_NAME/\//-} =~ ^v[0-9]+\.[0-9]+\.[0-9]+(-.+)?$ ]]; then
-          echo "VERSION=${GITHUB_REF_NAME/\//-}" >> $GITHUB_ENV
+          tag=${{GITHUB_REF_NAME}}
```
guicassolato, line 112: `GITHUB_REF_NAME` often contains slashes, such as in the case of branch names like `docs/foo`.

didierofrivia: yup, but it would be the same for the if condition, no?

guicassolato: I mean when we set the version to it, we need to replace all slashes with another char, as the slash symbol is not allowed in the manifests, I recon.

maleck13: so on 114 we need to add the string replace? 
`/\//-` ?

guicassolato: Or add it on 112 and remove the leading "v" in the regex in 113.

didierofrivia: I'd say it's not needed, since `VERSION` assignment happens within 2 branches of the conditional. First case matching the REGEX, SemVer with leading "v" and beta tags (as in `v0.9.0-cr1`) and the second when doesn't match it assigns the ref sha.

guicassolato: Understood, @didierofrivia. Thanks for clarifying.

guicassolato (approved):

PR #207, opened 2024-09-05
fix: ignore template diffs when verifying manifests

guicassolato:
Make the manifest/bundle verification to ignore changes to the authorino kustomization file. The changes to the manifests are the ones that actually matter, not the ones to the template/kustomization file.

Fixes: https://github.com/Kuadrant/authorino-operator/actions/runs/10716422467/job/29715385421


mikenairn:
Question: Why would we want to have manifests and a bundle on a branch/tag of this repo that do not have the correct image refs?

eguzki (approved):

Issue #98, opened 2022-09-05
`RuntimeDefault` seccomp profile in the SCCs


guicassolato:
Authorino Operator [v0.4.1](https://github.com/Kuadrant/authorino-operator/releases/tag/v0.4.1) had to drop `RuntimeDefault` seccomp profile in the security context for backward compatibility with older versions of OpenShift (#97).

Hopefully this can soon be added back,
1. because the minimum version of OpenShift supported for the Authorino Operator became 4.11+, or
2. because older versions of OpenShift based on compatible versions of Kubernetes for this feature (i.e. v1.22+) decided to address the issue and allow `RuntimeDefault` seccomp profile (see https://github.com/openshift/cluster-kube-apiserver-operator/issues/1325).

KevFan:
Do we want to add this back now considering it's been 2 years since 4.11 was released ? 🤔

PR #208, opened 2024-09-12
feat: use RuntimeDefault seccomp profile in security context

KevFan:
# Description
Closes: https://github.com/Kuadrant/authorino-operator/issues/98

* Re-introduces change to use `RuntimeDefault` as the seccomp profile
* Annotate the bundle for compatibility with OpenShift 4.12+ (https://github.com/redhat-openshift-ecosystem/community-operators-prod/commit/6849cac05456d1074c37bc62e431ea5157a44138)
  * Bundle is always regenerated so these labels were added through the makefile


KevFan:
Verified successfull install on CRC locally that is using openshift 4.16.7

image

guicassolato (approved):
Thanks @KevFan!

PR #212, opened 2024-09-17
feat: set quay image expiry to prevent overflow of images

KevFan:
# Description
Part of: https://github.com/Kuadrant/kuadrant-operator/issues/769
Associated with: https://github.com/Kuadrant/kuadrant-operator/pull/843

Currently, all built Quay images never expire and are consuming unnecessary quota on Quay.io. Quay provides a way to [set tag expiration from a Dockerfile](https://docs.redhat.com/en/documentation/red_hat_quay/3.4/html/use_red_hat_quay/working_with_tags#setting_tag_expiration_from_a_dockerfile) to prevent this issue for images that don't need to be kept indefinitely, such as development branches and nightly builds. This can help prevent the problem from escalating over time.

# Verification
Since there is no auto trigger of building the images on dev branches, you at manually verify at least the makefile changes by:

  *  Remove local image
     ```
     docker rmi quay.io/kuadrant/authorino-operator:latest
     docker rmi quay.io/kuadrant/authorino-operator-bundle:latest
     docker rmi quay.io/kuadrant/authorino-operator-catalog:latest
     ```
  * Build image
    ```
     # Delete the generated catalog Dockerfile as it won't regenerate again with the correct label if it's already present
     rm catalog/authorino-operator-catalog.Dockerfile 
     make docker-build bundle-build catalog-dockerfile catalog-build
    ```
  * Inspect image labels 
    ```
    docker inspect quay.io/kuadrant/authorino-operator:latest --format='{{json .Config.Labels}}'
    docker inspect quay.io/kuadrant/authorino-operator-bundle:latest --format='{{json .Config.Labels}}'
    docker inspect quay.io/kuadrant/authorino-operator-catalog:latest --format='{{json .Config.Labels}}'
    ```
  * The `quay.expires-after` label should be set to `never`
     
    image

  * Remove local image again
     ```
     docker rmi quay.io/kuadrant/authorino-operator:latest
     docker rmi quay.io/kuadrant/authorino-operator-bundle:latest
     docker rmi quay.io/kuadrant/authorino-operator-catalog:latest
     ```
  * Build image with a set expiry
    ```
     # Delete the generated catalog Dockerfile as it won't regenerate again with the correct label if it's already present
     rm catalog/authorino-operator-catalog.Dockerfile 
     QUAY_IMAGE_EXPIRY=3d make docker-build bundle-build catalog-dockerfile catalog-build
    ```
  * Inspect image labels 
    ```
    docker inspect quay.io/kuadrant/authorino-operator:latest --format='{{json .Config.Labels}}'
    docker inspect quay.io/kuadrant/authorino-operator-bundle:latest --format='{{json .Config.Labels}}'
    docker inspect quay.io/kuadrant/authorino-operator-catalog:latest --format='{{json .Config.Labels}}'
    ```
  * The `quay.expires-after` label should be set to `3d`
    
     image

### DIFF
--- a/.github/workflows/build-images-base.yaml
+++ b/.github/workflows/build-images-base.yaml
@@ -1,10 +1,20 @@
 name: Build and push images
 
 on:
-  push:
-    branches:
-      - 'main'
-      - 'master'
+  workflow_call:
+    inputs:
+      authorinoVersion:
+        description: Authorino version
+        required: true
+        default: latest
+      channels:
+        description: Bundle and catalog channels, comma separated
+        required: true
+        default: stable
+      quayImageExpiry:
+        description: When to expire the built quay images. The time values could be something like 1h, 2d, 3w for hours, days, and weeks, respectively, from the time the image is built.
+        default: never
+        type: string
   workflow_dispatch:
     inputs:
       authorinoVersion:
@@ -15,15 +25,19 @@ on:
         description: Bundle and catalog channels, comma separated
         required: true
         default: stable
+      quayImageExpiry:
+        description: When to expire the built quay images. The time values could be something like 1h, 2d, 3w for hours, days, and weeks, respectively, from the time the image is built.
+        default: never
+        type: string
 
 env:
-  IMG_TAGS: ${{ github.sha }}
+  IMG_TAGS: ${{ inputs.authorinoVersion }}
   IMG_REGISTRY_HOST: quay.io
   IMG_REGISTRY_ORG: kuadrant
-  MAIN_BRANCH_NAME: main
   OPERATOR_NAME: authorino-operator
   BUILD_CONFIG_FILE: build.yaml
   LATEST_AUTHORINO_GITREF: ${{ vars.AUTHORINO_SHA != '' && vars.AUTHORINO_SHA || 'latest' }}
+  QUAY_IMAGE_EXPIRY: ${{ inputs.quayImageExpiry }}
 
 jobs:
   build:
@@ -32,16 +46,6 @@ jobs:
     steps:
       - name: Check out code
         uses: actions/checkout@v3
-      - name: Add latest tag
-        if: ${{ github.ref_name == env.MAIN_BRANCH_NAME }}
-        id: add-latest-tag
-        run: |
-          echo "IMG_TAGS=latest ${{ env.IMG_TAGS }}" >> $GITHUB_ENV
-      - name: Add branch tag
-        if: ${{ github.ref_name != env.MAIN_BRANCH_NAME }}
-        id: add-branch-tag
-        run: |
-          echo "IMG_TAGS=${GITHUB_REF_NAME/\//-} ${{ env.IMG_TAGS }}" >> $GITHUB_ENV
       - name: Set Operator version
         id: operator-version
         run: |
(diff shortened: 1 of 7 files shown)


.github/workflows/build-images-main-sha.yaml
```diff
@@ -0,0 +1,14 @@
...
+    branches: ['main']
+
+jobs:
+  workflow-build:
+    name: Calls build-images-base workflow
+    uses: ./.github/workflows/build-images-base.yaml
+    secrets: inherit
+    with:
+      authorinoVersion: ${{ github.sha }}
+      quayImageExpiry: 2w
```
KevFan, line 14: I'm making an assumption that SHA build are equivalent to nightly builds that we want to expire 🤔 

If anyone has a better expiry value, I can update this

.github/workflows/build-images-base.yaml
```diff
@@ -32,16 +46,6 @@ jobs:
...
-      - name: Add latest tag
-        if: ${{ github.ref_name == env.MAIN_BRANCH_NAME }}
-        id: add-latest-tag
-        run: |
-          echo "IMG_TAGS=latest ${{ env.IMG_TAGS }}" >> $GITHUB_ENV
-      - name: Add branch tag
-        if: ${{ github.ref_name != env.MAIN_BRANCH_NAME }}
-        id: add-branch-tag
-        run: |
-          echo "IMG_TAGS=${GITHUB_REF_NAME/\//-} ${{ env.IMG_TAGS }}" >> $GITHUB_ENV
```
guicassolato, line 44: Is this really what we want? This will break with the single manifest link between builds.

.github/workflows/build-images-main-sha.yaml
guicassolato: Until now, SHA builds have not been equivalent to nightly builds. As of today, builds are tagged _additionally_ with the commit SHA on top of the other, more human-friendly tag. This includes `latest` (for `main` branch), feature branches, and long-lived release builds.

I'm not sure how I feel about breaking this link between tags, effectively building two separate images now, if this is only for the purpose of cleaning up the registry.

The `quay.expires-after` label seems easy enough, but it only works for expiring an entire image, not an individual image tag.

guicassolato (changes requested):
I think we should think this through a little more. Breaking the link between images tags into the single unified manifests they are currently build today does not seem the correct approach to solve the problem of cleaning up the registry.

.github/workflows/build-images-base.yaml
KevFan: I don't see how we can add the expire label for different types of builds unless we do this since latest & SHA builds are usually the same on merge to main. 

Branch + SHA build would only happen if the workflow is triggered manually currently.

.github/workflows/build-images-main-sha.yaml
KevFan: Hmm, I guess the question is then, does authorino-operator images need / want to expire images through this label? To me, it seems like it may not since the current workflow triggers on merge to main, which would build with an equivalent SHA tag. These SHA tags are kept indefinitely, but maybe that is something we want.

Feature branch and release builds looks to require manual trigger of the same workflow which do not happen that regularly

KevFan:
I don't think we necessarily want this, so I'm going to close this and this can be revisited if this is something that people want done

PR #229, opened 2024-11-14
RELATED_IMAGE_AUTHORINO env var

eguzki:
### What

Authorino image can be read from `RELATED_IMAGE_AUTHORINO` env var. Defaults to `quay.io/kuadrant/authorino:latest`

Required to specify the authorino image from the CSV manifest

The authorino image being used in the deployment is read from different sources with some order of precedence:
1. Authorino CR's `spec.image`
1. `RELATED_IMAGE_AUTHORINO` env var
1. Value set at  the operator  **compilation time** passed as ldflags `-X github.com/kuadrant/authorino-operator/controllers.DefaultAuthorinoImage=VALUE`
1. `quay.io/kuadrant/limitador:latest`
 
### Verification steps

run cluster
```
kind create cluster --name authorino-operator-system
```

install CRDs
```
make install
```

run operator locally with specific `RELATED_IMAGE_AUTHORINO` env var.

```
RELATED_IMAGE_AUTHORINO=quay.io/kuadrant/authorino:v0.18.0 make run
```

Create authorino object without `spec.image` to deploy version from env var.
```yaml
k apply -f - <<EOF
kind: Authorino
metadata:
  name: authorino
spec:
  listener:
    tls:
      enabled: false
  oidcServer:
    tls:
      enabled: false
EOF
```

Check authorino's version is `quay.io/kuadrant/authorino:v0.18.0`
```sh
$ kubectl get deployments authorino -o jsonpath='{.spec.template.spec.containers[0].image}'
```

The response should be 

```
quay.io/kuadrant/authorino:v0.18.0
```

Clean up

```
kind delete cluster --name authorino-operator-system
```

### DIFF
--- a/controllers/authorino_controller.go
+++ b/controllers/authorino_controller.go
@@ -31,6 +31,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/env"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
@@ -191,8 +192,15 @@ func (r *AuthorinoReconciler) buildAuthorinoDeployment(authorino *api.Authorino)
 	var containers []k8score.Container
 	var saName = authorino.Name + "-authorino"
 
-	if authorino.Spec.Image == "" {
-		authorino.Spec.Image = DefaultAuthorinoImage
+	image := authorino.Spec.Image
+
+	if image == "" {
+		image = env.GetString("RELATED_IMAGE_AUTHORINO", DefaultAuthorinoImage)
+	}
+
+	if image == "" {
+		// `DefaultAuthorinoImage can be empty string. But image cannot be or deployment will fail
+		image = "quay.io/kuadrant/authorino:latest"
 	}
 
 	var volumes []k8score.Volume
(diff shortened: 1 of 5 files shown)


eguzki:
supersedes https://github.com/Kuadrant/authorino-operator/pull/174

controllers/authorino_controller.go
```diff
@@ -191,8 +192,15 @@ func (r *AuthorinoReconciler) buildAuthorinoDeployment(authorino *api.Authorino)
...
-		authorino.Spec.Image = DefaultAuthorinoImage
+	image := authorino.Spec.Image
+
+	if image == "" {
+		image = env.GetString("RELATED_IMAGE_AUTHORINO", DefaultAuthorinoImage)
+	}
+
+	if image == "" {
+		// `DefaultAuthorinoImage can be empty string. But image cannot be or deployment will fail
+		image = "quay.io/kuadrant/authorino:latest"
```
guicassolato, line 203: This is dangerous. `DefaultAuthorinoImage` should never be empty. If we're keeping this build-arg variable as an option to set the default Authorino image, then IMO the variable should be set at every build to a non-empty value.

If this variable can be empty, I think I'd rather break the operator than make it silently reset all Authorino instances in the cluster to `latest` if the `RELATED_IMAGE_AUTHORINO` is removed from the deployment.

guicassolato (changes requested):

controllers/authorino_controller.go
adam-cattermole, line 203: I guess we can't set the default value of `DefaultAuthorinoImage` to the image because it's not unset it's empty string in this case?

eguzki: exactly

eguzki: ok, panicking if it is empty

adam-cattermole (approved):
Assuming @guicassolato is happy with the proposed solution, I've verified this locally with the various different settings for `DEFAULT_AUTHORINO_IMAGE`, `.spec.image` and `RELATED_IMAGE_AUTHORINO` and it all seems to be working well (and following intended precedence)

eguzki:
he requested changes, so I ~cannot~ shoud not merge even with one approval.

guicassolato (approved):
I'm happy with panicking if it goes all the way down and `DefaultAuthorinoImage` is empty. Thanks for addressing it!

I also like the explanation about the order of precedence in the description of the PR. We should probably add it to the last paragraph of https://github.com/Kuadrant/authorino-operator/blob/main/RELEASE.md (otherwise `RELATED_IMAGE_AUTHORINO` in undocumented).

But I'll approve the PR now, in light that it has been tested and it works. Docs can be added later.

Thanks, @eguzki and @adam-cattermole!

PR #233, opened 2024-11-21
fix: verify-manifest/bundle

guicassolato:


alexsnaps (approved):
🫣

Makefile
```diff
@@ -354,20 +354,25 @@ prepare-release: ## Prepares a release: create build info file, generate manifes
 
 ## Targets to verify actions that generate/modify code have been executed and output committed
 
+OPERATOR_IMAGE_REPO = $(shell echo $(OPERATOR_IMAGE) | cut -d: -f1)
+DEFAULT_AUTHORINO_IMAGE_REPO = $(shell echo $(DEFAULT_AUTHORINO_IMAGE) | cut -d: -f1)
+
 .PHONY: verify-manifests
 verify-manifests: manifests $(YQ) ## Verify manifests update.
-	git diff -I'^    createdAt:' -I' containerImage:' -I' image:' --exit-code -- ./config ':(exclude)config/authorino/kustomization.yaml'
+	git diff -I'^    createdAt:' -I'$(OPERATOR_IMAGE_REPO)' -I'$(DEFAULT_AUTHORINO_IMAGE_REPO)' -- ./config ':(exclude)config/authorino/kustomization.yaml'
```
eguzki, line 362: removing `--exit-code` ? it will not fail if there is some change

guicassolato: `--exit-code` makes it exist with 1 if there is any difference in the file (other than the ones ignored of course) and 0 otherwise. We want that.

eguzki: Confused here. When you say "we want that", you mean we want `--exit-code` ? Then why did you remove it?

guicassolato: I did!!?? Damn! I thought you were asking to remove it and I argued we need it, but didn’t realise the mistake. Fixing it now.

Thanks @eguzki

guicassolato: https://github.com/Kuadrant/authorino-operator/pull/234

PR #237, opened 2025-01-09
chore(test): simplifies ginkgo/gomega tests

bartoszmajsak:
This PR simplifies few things in ginkgo/gomega tests:

- uses ctx propagation instead of TODO
- sets default timeouts and interval for `Eventually` checks eliminating
  duplication
- reworks some assertions to make them easier to read

Signed-off-by: bartoszmajsak <bartosz.majsak@gmail.com>

### DIFF
--- a/controllers/authorino_controller_test.go
+++ b/controllers/authorino_controller_test.go
@@ -10,17 +10,18 @@ import (
 	k8score "k8s.io/api/core/v1"
 	k8srbac "k8s.io/api/rbac/v1"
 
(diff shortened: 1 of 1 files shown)


bartoszmajsak:
Not quite sure what to do with GitGuardian - it's an existing code and presumably made-up hash. Is there a way to mark it as false-positive?

KevFan (approved):
Nice! Looks good to me 👍

PR #240, opened 2025-02-03
 [ArtifactHub.io] GPG public key annotation added

didierofrivia:
Part of the work needed for https://github.com/Kuadrant/helm-charts/issues/18

This PR adds a new annotation in the Helm Chart.yaml meant for displaying the public key used for signing its chart packages. More info in https://artifacthub.io/docs/topics/annotations/helm/


guicassolato (approved):
I've downloaded the key hosted at the URL myself and confirmed the fingerprint matches.

Thanks @didierofrivia

PR #243, opened 2025-02-11
fix(labels): do not remove user added labels from Deployment's spec.template

### LINKED ISSUE #239: Do not overwrite user added labels from deployment spec.template.labels in order to propagate the labels to the pod
**Is your feature request related to a problem? Please describe.**
Expose configuration in the authorino cr which allows a user to ask for the addition of the following lable to the authorino deployment `"sidecar.istio.io/inject":"true"`

**Describe the solution you'd like**
A value exposed in the cr which the operator responds to by applying the label. 
The label should be applied if the value is set to true
The label should be removed if the value isn't present or set to false. 

**Additional context**
This work forms part of a set of issues to enable the automatic configuration of https://github.com/Kuadrant/kuadrant-operator/blob/a1ca1d0ee2ca7dac79ad921b0ff30ded2c986612/doc/install/mtls-configuration.md#configure-mtls-between-the-gateway-and-kuadrant-components

laurafitzgerald:
PR #91 brought the ability to propagate labels defined for Authorino CR down to Deployment. However, more often than not, it's more desired to have those labels propagated to template and therefore pods which are part of the deployment. This can simplify e.g. making Authorino part of service mesh, as injection label can be consistently added to relevant resources.

~When creating the deployment, instead of having fixed set of labels for Deployment's `spec.template` and objectmeta.lables this change appends Authorino CR labels as well.~

EDIT:  This change will now allow the ability to add a label via your chosen mechanism to the deployment.spec.template.labels without being overwritten by the operator. 

As per discussion at https://groups.google.com/a/redhat.com/g/kuadrant-dev/c/rbsY4kDaSnE

Closes https://github.com/Kuadrant/authorino-operator/issues/239
Replaces https://github.com/Kuadrant/authorino-operator/pull/236

### DIFF
--- a/controllers/authorino_controller.go
+++ b/controllers/authorino_controller.go
@@ -66,13 +66,12 @@ type AuthorinoReconciler struct {
 // +kubebuilder:rbac:groups="",resources=events,verbs=create;patch;
 // +kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch
 // +kubebuilder:rbac:groups="authorino.kuadrant.io",resources=authconfigs,verbs=create;delete;get;list;patch;update;watch
-// +kubebuilder:rbac:groups="authorino.kuadrant.io",resources=authconfigs,verbs=create;delete;get;list;patch;update;watch
 // +kubebuilder:rbac:groups="authorino.kuadrant.io",resources=authconfigs/status,verbs=get;patch;update
 // +kubebuilder:rbac:groups="coordination.k8s.io",resources=leases,verbs=get;list;create;update;
 
 // Reconcile deploys an instance of authorino depending on the settings
-// defined in the API, any change applied to the existings CRs will trigger
-// a new reconcilation to apply the required changes
+// defined in the API, any change applied to the existing CRs will trigger
+// a new reconciliation to apply the required changes
 func (r *AuthorinoReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	logger := r.Log.WithValues("authorino", req.NamespacedName)
 
@@ -106,14 +105,21 @@ func (r *AuthorinoReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	}
 
 	// Gets Deployment resource for the authorino instance
-	if existingDeployment, err := r.getAuthorinoDeployment(authorinoInstance); err != nil {
+	existingDeployment, err := r.getAuthorinoDeployment(authorinoInstance)
+	if err != nil {
 		return ctrl.Result{}, r.wrapErrorWithStatusUpdate(logger, authorinoInstance, r.setStatusFailed(statusUnableToGetDeployment),
 			fmt.Errorf("failed to get %s Deployment resource, err: %v", authorinoInstance.Name, err),
 		)
-	} else if existingDeployment == nil {
+	}
+	if existingDeployment == nil {
 		// Creates a new deployment resource to deploy the new authorino instance
-		newDeployment := r.buildAuthorinoDeployment(authorinoInstance)
-		if err := r.Client.Create(context.TODO(), newDeployment); err != nil {
+		newDeployment, err := r.buildAuthorinoDeployment(authorinoInstance)
+		if err != nil {
+			return ctrl.Result{}, r.wrapErrorWithStatusUpdate(logger, authorinoInstance, r.setStatusFailed(statusUnableToBuildDeploymentObject),
+				fmt.Errorf("failed to build %s Deployment resource for creation, err: %v", authorinoInstance.Name, err),
+			)
+		}
+		if err := r.Client.Create(ctx, newDeployment); err != nil {
 			return ctrl.Result{}, r.wrapErrorWithStatusUpdate(
 				logger, authorinoInstance, r.setStatusFailed(statusUnableToCreateDeployment),
 				fmt.Errorf("failed to create %s Deployment resource, err: %v", newDeployment.Name, err),
@@ -126,13 +132,23 @@ func (r *AuthorinoReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		// Deployment created successfully - return and requeue
 		return ctrl.Result{Requeue: true}, nil
 	} else {
+
 		// deployment already exists, then build a new resource with the desired changes
 		// and compare them, if changes are encountered apply the desired changes
-		desiredDeployment := r.buildAuthorinoDeployment(authorinoInstance)
-		logger.Info("desiredDeployment", "deployment", desiredDeployment)
-		logger.Info("existingDeployment", "deployment", existingDeployment)
+		desiredDeployment, err := r.buildAuthorinoDeployment(authorinoInstance)
+		if err != nil {
+			return ctrl.Result{}, r.wrapErrorWithStatusUpdate(logger, authorinoInstance, r.setStatusFailed(statusUnableToBuildDeploymentObject),
+				fmt.Errorf("failed to build %s Deployment resource for updating, err: %v", authorinoInstance.Name, err),
+			)
+		}
+		logger.V(1).Info("desiredDeployment", "deployment", desiredDeployment)
+		logger.V(1).Info("existingDeployment", "deployment", existingDeployment)
+		// merge existing key/value pairs back into the desiredDeployment Spec Template Labels so as not to overwrite those.
+		for existingKey, existingValue := range existingDeployment.Spec.Template.Labels {
+			desiredDeployment.Spec.Template.Labels[existingKey] = existingValue
+		}
 		if changed := r.authorinoDeploymentChanges(existingDeployment, desiredDeployment); changed {
-			if err := r.Update(ctx, desiredDeployment); err != nil {
+			if err = r.Update(ctx, desiredDeployment); err != nil {
 				return ctrl.Result{}, r.wrapErrorWithStatusUpdate(
 					logger, authorinoInstance, r.setStatusFailed(statusUnableToUpdateDeployment),
 					fmt.Errorf("failed to update %s Deployment resource, err: %v", desiredDeployment.Name, err),
@@ -188,7 +204,7 @@ func (r *AuthorinoReconciler) getAuthorinoDeployment(authorino *api.Authorino) (
 	return deployment, nil
 }
 
-func (r *AuthorinoReconciler) buildAuthorinoDeployment(authorino *api.Authorino) *k8sapps.Deployment {
+func (r *AuthorinoReconciler) buildAuthorinoDeployment(authorino *api.Authorino) (*k8sapps.Deployment, error) {
 	var containers []k8score.Container
 	var saName = authorino.Name + "-authorino"
 
@@ -259,7 +275,7 @@ func (r *AuthorinoReconciler) buildAuthorinoDeployment(authorino *api.Authorino)
 		volumes = append(volumes, authorinoResources.GetTlsVolume(authorinoTlsCertVolumeName, secretName))
 	}
 
-	// if an external OIDC server is enable mounts a volume to the container
+	// if an external OIDC server is enabled mounts a volume to the container
 	// by using the secret with the certs
 	if enabled := authorino.Spec.OIDCServer.Tls.Enabled; enabled == nil || *enabled {
 		secretName := authorino.Spec.OIDCServer.Tls.CertSecret.Name
@@ -270,7 +286,7 @@ func (r *AuthorinoReconciler) buildAuthorinoDeployment(authorino *api.Authorino)
 	args := r.buildAuthorinoArgs(authorino)
 	var envs []k8score.EnvVar
 
-	// [DEPRECATED] configure authorino using env vars (only for old Authorino versions)
+	// Deprecated: configure authorino using env vars (only for old Authorino versions)
 	authorinoVersion := authorinoVersionFromImageTag(image)
 	if detectEnvVarAuthorinoVersion(authorinoVersion) {
 		envs = r.buildAuthorinoEnv(authorino)
@@ -308,8 +324,8 @@ func (r *AuthorinoReconciler) buildAuthorinoDeployment(authorino *api.Authorino)
 		authorino.Labels,
 	)
 
-	_ = ctrl.SetControllerReference(authorino, deployment, r.Scheme)
-	return deployment
+	err := ctrl.SetControllerReference(authorino, deployment, r.Scheme)
+	return deployment, err
 }
 
 func (r *AuthorinoReconciler) buildAuthorinoArgs(authorino *api.Authorino) []string {
@@ -425,7 +441,7 @@ func (r *AuthorinoReconciler) buildAuthorinoArgs(authorino *api.Authorino) []str
 	return args
 }
 
-// [DEPRECATED] Configures Authorino by defining environment variables (instead of command-line args)
+// Deprecated: Configures Authorino by defining environment variables (instead of command-line args)
 // Kept for backward compatibility with older versions of Authorino (<= v0.10.x)
 func (r *AuthorinoReconciler) buildAuthorinoEnv(authorino *api.Authorino) []k8score.EnvVar {
 	envVar := []k8score.EnvVar{}
@@ -551,6 +567,16 @@ func (r *AuthorinoReconciler) buildAuthorinoEnv(authorino *api.Authorino) []k8sc
 
 func (r *AuthorinoReconciler) authorinoDeploymentChanges(existingDeployment, desiredDeployment *k8sapps.Deployment) bool {
 	if *existingDeployment.Spec.Replicas != *desiredDeployment.Spec.Replicas {
+		return true
+	}
+
+	// check the labels here to see if there needs to be a change
+	// selector labels are immutable so don't need to check those.
+	if authorinoResources.MapUpdateNeeded(existingDeployment.Labels, desiredDeployment.Labels) {
+		return true
+	}
+
+	if authorinoResources.MapUpdateNeeded(existingDeployment.Spec.Template.Labels, desiredDeployment.Spec.Template.Labels) {
 		return true
 	}
 
--- a/controllers/authorino_controller_test.go
+++ b/controllers/authorino_controller_test.go
@@ -3,6 +3,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"k8s.io/utils/env"
 	"strings"
 	"time"
 
@@ -54,14 +55,6 @@ var _ = Describe("Authorino controller", func() {
 
 			authorinoInstance = newFullAuthorinoInstance()
 			Expect(k8sClient.Create(ctx, authorinoInstance)).To(Succeed())
-
-			nsdName := namespacedName(authorinoInstance.GetNamespace(), authorinoInstance.GetName())
-
-			Eventually(func(ctx context.Context) bool {
-				var authorino api.Authorino
-				err := k8sClient.Get(ctx, nsdName, &authorino)
-				return err == nil && authorinoInstance.Status.Ready()
-			}).WithContext(ctx).Should(BeFalse())
 		})
 
 		It("Should create authorino required services", func(ctx context.Context) {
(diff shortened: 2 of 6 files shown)


bundle/manifests/authorino.kuadrant.io_authconfigs.yaml
```diff
@@ -72,7 +72,7 @@ spec:
             type: object
           spec:
             description: Specifies the desired state of the AuthConfig resource, i.e.
-              the authencation/authorization scheme to be applied to protect the matching
+              the authentication/authorization scheme to be applied to protect the matching
```
KevFan, line 75: I think the failing CI is due to this should be updated at the CRD at **authorino** instead as a separate PR:
- https://github.com/Kuadrant/authorino/blob/b2bde91bfc64b1d140f740069ba36ce81a3f3349/api/v1beta2/auth_config_types.go#L109
- https://github.com/Kuadrant/authorino/blob/b2bde91bfc64b1d140f740069ba36ce81a3f3349/api/v1beta3/auth_config_types.go#L110

And the manifests can be updated using `make manifests generate` in **authorino**. 

**Authorino operator** can then update it's own bundle/manifest/chart using it's own make command without the CI failing one that has merged 🤔

bundle/manifests/authorino.kuadrant.io_authconfigs.yaml
```diff
@@ -2606,8 +2606,13 @@ spec:
             type: object
           spec:
             description: Specifies the desired state of the AuthConfig resource, i.e.
+<<<<<<< HEAD
```
guicassolato, line 2609: Some conflict resolution leftovers here.

pkg/resources/k8s_deployment.go
```diff
@@ -75,3 +75,16 @@ func GetTlsVolume(certName, secretName string) k8score.Volume {
 		},
 	}
 }
+
+func MapUpdateNeeded(existing map[string]string, desired map[string]string) bool {
+	if existing == nil {
```
eguzki, line 80: no need to check if it's null. it's ok to check if a key exists in a nil map.

controllers/authorino_controller.go
```diff
@@ -126,11 +132,29 @@ func (r *AuthorinoReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+			err = updateStatusConditions(logger, authorinoInstance, r.Client, statusNotReady(statusDeploymentNotReady, "Authorino Deployment resource not ready"))
+			return ctrl.Result{RequeueAfter: time.Second}, err
+		}
+
 		// deployment already exists, then build a new resource with the desired changes
 		// and compare them, if changes are encountered apply the desired changes
-		desiredDeployment := r.buildAuthorinoDeployment(authorinoInstance)
-		logger.Info("desiredDeployment", "deployment", desiredDeployment)
-		logger.Info("existingDeployment", "deployment", existingDeployment)
+		desiredDeployment, err := r.buildAuthorinoDeployment(authorinoInstance)
```
eguzki, line 144: I would check the value of the error right after the call

controllers/authorino_controller.go
```diff
@@ -126,11 +132,29 @@ func (r *AuthorinoReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+		// merge existing key/value pairs back into the desiredDeployment Spec Template Labels so as not to overwrite those.
+		for existingKey, existingValue := range existingDeployment.Spec.Template.Labels {
+			if desiredValue, exists := (desiredDeployment.Spec.Template.Labels)[existingKey]; !exists || desiredValue != existingValue {
```
eguzki, line 147: what if the user adds a label which key conflicts with desired one? for instance `authorino-resource: something else` ?

controllers/authorino_controller.go
```diff
@@ -126,11 +132,29 @@ func (r *AuthorinoReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+				desiredDeployment.Spec.Template.Labels[existingKey] = existingValue
+			}
+		}
+		if err != nil {
+			return ctrl.Result{}, r.wrapErrorWithStatusUpdate(logger, authorinoInstance, r.setStatusFailed(statusUnableToBuildDeploymentObject),
+				fmt.Errorf("failed to build %s Deployment resource for updating, err: %v", authorinoInstance.Name, err),
+			)
+		}
+		logger.V(1).Info("desiredDeployment", "deployment", desiredDeployment)
```
eguzki, line 156: This logging happens after potentially mutating the desired deployment. Is that intended? I would log (for dev/testing purposes as it is in debug mode) the existing and desired as they are without any update.

eguzki (commented):
Looks good. I left some comments to discuss

controllers/authorino_controller.go
laurafitzgerald: Can this happen on update? the deployment would already be created with the existing hard coded ones. so the only edit they could make is to the value. if they do that we want to set it back to the desired value right?

controllers/authorino_controller.go
```diff
@@ -126,11 +132,29 @@ func (r *AuthorinoReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		// Deployment created successfully - return and requeue
 		return ctrl.Result{Requeue: true}, nil
 	} else {
+
+		if !deploymentAvailable(existingDeployment) {
```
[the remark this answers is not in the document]
laurafitzgerald: this change is just moving existing logic up to here. my thinking is that if the deployment is not available we should requeue here rather than going through the rest of the logic and then checking if it's not available. if it shouldn't be in this controller, I can move it. or if it's better to move it back to where it was, to prevent the deadlock i can do that.

controllers/authorino_controller.go
laurafitzgerald: yes, agreed, i'll move the log up for the desired deployment so that it's showing the desired without the user added labels. This will all be improved as well later when implementing the mutator pattern.

controllers/authorino_controller.go
eguzki: Right, it is not about _adding_ but updating the value. If that happens, the pod templates would not match the selector and API server would reject the call so you would get an error on the update call, so you would try and try endlessly. Try it, as I might be wrong. If I am not wrong, I would add the logic to protect some label keys against changes existing in the PR of the limitador operator

laurafitzgerald: The selector labels are spec.selector.LabelSelector.MatchLabels. these are not edited in this case. so any hardcoded desired key will be put back to the value we expect. only if the desiredKey doesn't exist will an update be made. otherwise the value stays the same as what's in desired, which is the hardcoded ones.

controllers/authorino_controller.go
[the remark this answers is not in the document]
laurafitzgerald: it's been moved back to where it was. This check does actually cause some unintended difficulties in envtest. because the deployment status never gets updated in that case due to the fact it's not a full running k8s environment, it blocked some of the tests from operating correctly. i've updated some of the logic to reflect that.

controllers/authorino_controller.go
```diff
@@ -126,13 +132,25 @@ func (r *AuthorinoReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+		if err != nil {
+			return ctrl.Result{}, r.wrapErrorWithStatusUpdate(logger, authorinoInstance, r.setStatusFailed(statusUnableToBuildDeploymentObject),
+				fmt.Errorf("failed to build %s Deployment resource for updating, err: %v", authorinoInstance.Name, err),
+			)
+		}
+		logger.V(1).Info("desiredDeployment", "deployment", desiredDeployment)
+		logger.V(1).Info("existingDeployment", "deployment", existingDeployment)
+		// merge existing key/value pairs back into the desiredDeployment Spec Template Labels so as not to overwrite those.
+		for existingKey, existingValue := range existingDeployment.Spec.Template.Labels {
+			if desiredValue, exists := (desiredDeployment.Spec.Template.Labels)[existingKey]; !exists || desiredValue != existingValue {
```
guicassolato, line 148: Do we need this condition? Can't we just iterate over `existing` and set them all into `desired`?

controllers/authorino_controller.go
guicassolato: Sorry. Late for the party at https://github.com/Kuadrant/authorino-operator/pull/243#discussion_r1971854555, but sounds like @eguzki has a point. We must ensure pod labels match the deployment selector.

pkg/resources/k8s_deployment.go
```diff
@@ -6,9 +6,9 @@ import (
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-func GetDeployment(name, namespace, saName string, replicas *int32, containers []k8score.Container, vol []k8score.Volume, labels map[string]string) *k8sapps.Deployment {
-	objMeta := getObjectMeta(namespace, name, labels)
-	authorinoLabels := labelsForAuthorino(name)
+func GetDeployment(name, namespace, saName string, replicas *int32, containers []k8score.Container, vol []k8score.Volume) *k8sapps.Deployment {
+	authorinoLabels := defaultAuthorinoLabels(name)
```
eguzki, line 10: There is an issue that was there before this PR, and your PR brought it to the surface. You cannot use the same map to populate metadata.labels, spec.selector and template.metadata.labels. Because maps are references, all those would share (are sharing now actually) the same map, so you change one of them and "magically" all of them are changed at the same time. And spec.selector cannot be changed, hence k8s API server complained. 

Build custom maps for each field so they can be safely be updated independently.

eguzki (changes requested):
Found an issue that needs to be fixed

pkg/resources/k8s_deployment.go
guicassolato: Nice catch, @eguzki! Also, if the selector cannot change, we need to make sure they override the labels for the template after the modification made by the user. I.e., users can add other labels, they can even change existing ones, as long as those labels are not used in the selector. Right?

controllers/authorino_controller.go
laurafitzgerald: yes, the idea was to not overwrite the desired hardcode values, but due to the fact that k8s api rejects a change to those hard coded ones because they wouldn't match the unmuatable selector.MatchLabels which are all set at creation time, i think it's safe to remove this.

eguzki:
### Verification Steps

* Run local cluster
```shell
kind create cluster --name authorino-local
```

* Checkout to this PR branch and deploy the operator
```shell
make docker-build OPERATOR_IMAGE=quay.io/kuadrant/authorino-operator:eguzki-custom-build-01
make docker-push OPERATOR_IMAGE=quay.io/kuadrant/authorino-operator:eguzki-custom-build-01
make install OPERATOR_IMAGE=quay.io/kuadrant/authorino-operator:eguzki-custom-build-01
make deploy OPERATOR_IMAGE=quay.io/kuadrant/authorino-operator:eguzki-custom-build-01
```
* Deploy authorino instance
```shell
k create namespace authorino

k apply -n authorino -f - <<EOF
kind: Authorino
metadata:
  name: authorino
spec:
  listener:
    tls:
      enabled: false
  oidcServer:
    tls:
      enabled: false
EOF
```
* Check labels
```
k get pods -n authorino -l authorino-resource=authorino -o yaml | yq e '.items[0].metadata.labels'
```

The result should be the hardcoded ones

```
authorino-resource: authorino
control-plane: controller-manager
pod-template-hash: 5b6f45cccb
```

* Add user defined labels to deployment pod template

```shell
kubectl patch deployment -n authorino authorino --type merge --patch '{"spec": {"template": {"metadata": {"labels": {"sidecar.istio.io/inject": "true"}}}}}'
```

* Check authorino pod labels
```shell
k get pods -n authorino -l authorino-resource=authorino -o yaml | yq e '.items[0].metadata.labels'
```

It should return
```
authorino-resource: authorino
control-plane: controller-manager
pod-template-hash: 5d7c6567fc
sidecar.istio.io/inject: "true"
```


* Update authorino image using the RELATED_IMAGE_AUTHORINO env var to some different image tag. The env var update triggers rolling out first the operator pod. 
Then, the new operator pod should reconcile the deployment of authorino, hence rolling out a new authorino deployment.
The new authorino pod should keep the user defined labels
```
kubectl set env deployment/authorino-operator -n authorino-operator RELATED_IMAGE_AUTHORINO=quay.io/kuadrant/authorino:cbc71830174cd997d6c6b881ba5d9d94ba72a77f
```

Check that authorino operator env var has been updated:
```
k get deployment authorino-operator -n authorino-operator -o jsonpath='{.spec.template.spec.containers[0].env}'
```
The new image should be there:

```
[{"name":"RELATED_IMAGE_AUTHORINO","value":"quay.io/kuadrant/authorino:cbc71830174cd997d6c6b881ba5d9d94ba72a77f"}]
```
Check that authorino instance is running from the new image:
```
k get deployment authorino -n authorino -o jsonpath='{.spec.template.spec.containers[0].image}'
```
The new image should be:
```
quay.io/kuadrant/authorino:cbc71830174cd997d6c6b881ba5d9d94ba72a77f
```
Check that the new authorino instance still has the user defined labels

```shell
k get pods -n authorino -l authorino-resource=authorino -o yaml | yq e '.items[0].metadata.labels'
```

It should return
```
authorino-resource: authorino
control-plane: controller-manager
pod-template-hash: 5d7c6567fc
sidecar.istio.io/inject: "true"
```

* Clean up
```shell
kind delete cluster --name authorino-local
```

pkg/resources/k8s_deployment.go
laurafitzgerald: if they try to edit the existing values, which have a matching key in the selector, the k8s api will reject that change. i saw it happen in a test where i tried to update that to a bad value. the api rejected the call with error `type filevalueinvalid selector does not match template labels on field spec.template.metadata.labels`

lphiri:
@laurafitzgerald  - I  don't have full context, but am I understanding correctly that this does not propagate labels set on the Authorino CR as was proposed in the original #236?

eguzki:
> @laurafitzgerald - I don't have full context, but am I understanding correctly that this does not propagate labels set on the Authorino CR as was proposed in the original #236?

You are correct. The user needs to add the labels to deployment pod template. 

I would also have done the label propagation. 1) Because the user interface is the CR and 2) the deployment of authorino is owned by the operator and should not be updated externally.  On the community call, it was decided to do the way this PR implements because if done in the CR, it needs to be decided where to propagate the labels: i.e. to the deployment labels?, pod template labels? services? configmaps? I will keep pushing to have labels propagated from the authorino CR ;)

laurafitzgerald:
> @laurafitzgerald - I don't have full context, but am I understanding correctly that this does not propagate labels set on the Authorino CR as was proposed in the original #236?

Correct. This change allows the user to add the label themselves manually or via some automation. We have spoken with some users who are looking for this change who are fine with that. As @eguzki mentioned, we decided to not propagate the labels everywhere for now because the label may have unintended consequences on some of the other k8s resources. This may happen in the future or alternatively there may be some option added in the spec of the cr to do that. For now this change allows the users to add the required label and authorino-operator wont remove it. Please let us know if this does not satisfy your needs.

laurafitzgerald:
> 
> * Clean up
> 
> ```shell
> kind delete cluster --name authorino-local
> ```

@eguzki 
Just confirming the verification works on my side.  👍 
I've also tried editing the value of the existing hardcoded labels and the api rejects it with the error `selector` does not match template `labels`

zdtsw:
> > @laurafitzgerald - I don't have full context, but am I understanding correctly that this does not propagate labels set on the Authorino CR as was proposed in the original #236?
> 
> Correct. This change allows the user to add the label themselves manually or via some automation. We have spoken with some users (including @bartoszmajsak who opened 236) who are looking for this change who are fine with that. As @eguzki mentioned, we decided to not propagate the labels everywhere for now because the label may have unintended consequences on some of the other k8s resources. This may happen in the future or alternatively there may be some option added in the spec of the cr to do that. For now this change allows the users to add the required label and authorino-operator wont remove it. Please let us know if this does not satisfy your needs.

Please help me walk thro some scenarios:
if we (another operator)  would like to use OSSM envoyproxy, we will need to
- create Authorino CR
- with or without label on Authorino CR wont have any difference
- wait for the deployment of authorino gets created by Authornio Operator,  then we need to manually add `sidecar.istio.io/inject: "true"` to .`spec.template.metadata.labels`
- this label will eventually populated to Authorino pod
- Next time when end user upgrade to a newer Authorino release. deployment of authorino retains with this label, so as the pod.
This change of "label" applies to any operations , e.g 
- manual add new one, 
- update existin one, 
- remove old one. 
which means, accidently modify label to  `sidecar.istio.io/inject: "false"` wont get automatically rectified unless someone ( we as another operator or end user) correct it on authorino deployment.

KevFan (commented):
This looks good to me anyway 👍 I'll leave the approval to @eguzki  

When we do the next release of authorino operator containing this change, it's probably a good idea to put in the release notes that this change also **removes** propagating Authorino CR labels to the authorino deployment. While this feature was not advertised in our docs, it's a behaviour change for existing users, so I think it would be good to note this 📝

laurafitzgerald:
> > > @laurafitzgerald - I don't have full context, but am I understanding correctly that this does not propagate labels set on the Authorino CR as was proposed in the original #236?
> > 
> > 
> > Correct. This change allows the user to add the label themselves manually or via some automation. We have spoken with some users (including @bartoszmajsak who opened 236) who are looking for this change who are fine with that. As @eguzki mentioned, we decided to not propagate the labels everywhere for now because the label may have unintended consequences on some of the other k8s resources. This may happen in the future or alternatively there may be some option added in the spec of the cr to do that. For now this change allows the users to add the required label and authorino-operator wont remove it. Please let us know if this does not satisfy your needs.
> 
> Please help me walk thro some scenarios: if we (another operator) would like to use OSSM envoyproxy, we will need to
> 
> * create Authorino CR
> * with or without label on Authorino CR wont have any difference
> * wait for the deployment of authorino gets created by Authornio Operator,  then we need to manually add `sidecar.istio.io/inject: "true"` to .`spec.template.metadata.labels`
> * this label will eventually populated to Authorino pod
> * Next time when end user upgrade to a newer Authorino release. deployment of authorino retains with this label, so as the pod.

Yes, correct to this point.
As the label has been added by something external to authorino-operator if any of the below changes are made authorino won't do anything. It's basically unopinionated about any label added to spec.template.labels by something other than itself. So it would be down to something external to ensure that it is maintained as expected/needed. that would be the case if you were adding the label to the cr. you would need to ensure that is in a state that you want. so no difference in expectation there due to the fact that we are doing it this way.

>   This change of "label" applies to any operations , e.g
> * manual add new one,
> * update existin one,
> * remove old one.
>   which means, accidently modify label to  `sidecar.istio.io/inject: "false"` wont get automatically rectified unless someone ( we as another operator or end user) correct it on authorino deployment.
>   same to the case, if deployment gets crashloop or deleted, someone will need to re-apply label again.

lphiri:
@laurafitzgerald  @eguzki  @guicassolato in terms of  (longer term) requirements, I would say our real need is a declarative API to indicate that we want the Authorino deployment to be part of  a service mesh (with options to specify which mesh: OSSM, vanilla Istio etc) , so making that an explicit (optional)  config in the Authorino CR would be ideal, instead of the indirect label based approach.  IMHO its reasonable to expect this as part of the API given Authorino is going be commonly used with envoy based service meshes.

guicassolato:
> @laurafitzgerald @eguzki @guicassolato in terms of (longer term) requirements, I would say our real need is a declarative API to indicate that we want the Authorino deployment to be part of a service mesh (with options to specify which mesh: OSSM, vanilla Istio etc) , so making that an explicit (optional) config in the Authorino CR would be ideal, instead of the indirect label based approach. IMHO its reasonable to expect this as part of the API given Authorino is going be commonly used with envoy based service meshes.

@lphiri, I see it differently actually. Authorino does not know anything about the client that connects to it. It can be integrated from an Envoy-based proxy, but also from pretty much anything else. It doesn't know it's a gateway, a mesh proxy, Kube's API server (we also have written use cases for this), a client app directly, or anything at all. And I think it should remain ignorant about it.

If we want to support propagating a label from the Authorino CR, it should be a generic label, but nothing specific about mesh or sidecar injection.

I think the reason why this propagation was kept out of the initial design was because we want to be very surgical about which exact internal objects we propagate the labels to (Service, Deployment, Pod, etc). I hope we can incorporate this second part in a future PR.

lphiri:
@guicassolato I see. That perspective also makes sense.  
Will wait for the future PR - setting labels in the Authorino CR would certainly be better than what we are having to do now.

eguzki (commented):
Changes looks good to me. Before approving, I would like to clarify that we are all ok with removing the label propagation from the Authorino CR to the deployment object (not the deployment pod template field) as pointed out correctly by @KevFan 

If we are ok with that, LGTM

Verification steps worked :green_circle:

controllers/authorino_controller_test.go
```diff
@@ -130,13 +129,23 @@ var _ = Describe("Authorino controller", func() {
 			}).WithContext(ctx).Should(Succeed())
 
 			replicas := int32(testAuthorinoReplicas)
-			image := DefaultAuthorinoImage
+			image := authorinoInstance.Spec.Image
 			existContainer := false
 
-			Expect(deployment.Spec.Replicas).Should(Equal(&replicas))
-			Expect(deployment.Labels).Should(Equal(map[string]string{"thisLabel": "willPropagate"}))
```
KevFan, line 137: Same, we should continue to test for this label now that we are keeping the functionality

KevFan (commented):
Verified changes and verified Authorino CR labels -> Deployment labels propagation still works. Left minor comment regarding removed lines in tests that I think we should keep, but otherwise looks good 👍

eguzki (approved):

KevFan (approved):
Nice! Looks good to me 👍

PR #248, opened 2025-02-27
make run: debug level and development mode

eguzki:


guicassolato (approved):

eguzki:
rebasing here with a commit having verified signature

PR #256, opened 2025-04-14
🔨 Fix pod template labels mutator 🔨

eguzki:
### What

When the existing deployment template label has few extra labels, the pod template label mutator was returning `true` to update the resource even when there was no update to be done. The extra few labels are preserved by the mutator.  That apparently little issue of reporting "update" when it should not had secondary effects.  The status controller would never set the status as Ready because it is being told that there is an update still going on. 

### Verifications steps

* Checkout this branch

* Run cluster
```
kind create cluster --name authorino-operator-system
```

* Install CRDs
```
make install
```

* Run operator locally 

```
make run
```

* Create authorino object.
```yaml
k apply -f - <<EOF
kind: Authorino
metadata:
  name: authorino
spec:
  listener:
    tls:
      enabled: false
  oidcServer:
    tls:
      enabled: false
EOF
```

* Wait for authorino instance to be ready
```shell
kubectl wait --timeout=300s --for=condition=Ready authorino authorino
```

* Add manually few extra pod template labels
```shell
kubectl patch deployment authorino --type=merge --patch '{"spec": {"template": {"metadata": {"labels": {"foo": "bar", "foo2": "bar2"}}}}}'
```
* Wait for authorino instance to be ready
```shell
kubectl wait --timeout=300s --for=condition=Ready authorino authorino
```

If authorino CR gets the `Ready` condition in status, the fix worked. If trying out the same verification steps with current HEAD of `main`, the authorino status looks like this:
```yaml
❯ k get authorino authorino -o jsonpath="{.status}" | yq e -P
conditions:
  - lastTransitionTime: "2025-04-14T17:19:18Z"
    message: Authorino Deployment resource updated
    reason: Updated
    status: "False"
    type: Ready
```

### DIFF
--- a/pkg/resources/map_utils.go
+++ b/pkg/resources/map_utils.go
@@ -1,0 +1,22 @@
+package resources
+
+func MergeMapStringString(existing *map[string]string, desired map[string]string) bool {
+	if existing == nil {
+		return false
+	}
+	if *existing == nil {
+		*existing = map[string]string{}
+	}
+
+	// for each desired key value set, e.g. labels
+	// check if it's present in existing. if not add it to existing.
+	// e.g. preserving existing labels while adding those that are in the desired set.
+	modified := false
+	for desiredKey, desiredValue := range desired {
+		if existingValue, exists := (*existing)[desiredKey]; !exists || existingValue != desiredValue {
+			(*existing)[desiredKey] = desiredValue
+			modified = true
+		}
+	}
+	return modified
+}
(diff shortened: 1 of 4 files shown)


pkg/resources/map_utils.go
```diff
@@ -0,0 +1,22 @@
+package resources
+
+func MergeMapStringString(existing *map[string]string, desired map[string]string) bool {
```
maleck13, line 3: nit: might be nice to add a quick comment so its known the bool returned is to let you know if the map was modified?

eguzki: done

maleck13 (approved):

PR #253, opened 2025-03-31
Addition of mutators to authorino reconciler

Patryk-Stefanski:
Closes https://github.com/Kuadrant/authorino-operator/issues/247

PR #258, opened 2025-04-15
Refactor several reconcilers

eguzki:
### What

- [x] Added missing imagepullpolicy mutator (regression from https://github.com/Kuadrant/authorino-operator/pull/253)
- [x] Added missing container args mutator (regression from https://github.com/Kuadrant/authorino-operator/pull/253)
- [x] Fix deployment mutator for volumes and volumemounts. `deepequal` cannot be used as created value is not the same as read back, the cluster adds `defaultMode` attribute.

Some reconcilers were reporting NOTFOUND errors when they should not. Example:
```
2025-04-16T16:33:32+02:00	ERROR	authorino-operator.controller.Authorino	Reconciler error	{"controller": "authorino", "controllerGroup": "operator.authorino.kuadrant.io", "controllerKind": "Authorino", "Authorino": {"name":"authorino","namespace":"default"}, "namespace": "default", "name": "authorino", "reconcileID": "4cf26c02-8a88-4325-ae02-b89f6954f707", "error": "failed to get authorino-leader-election-role Role, err: Role.rbac.authorization.k8s.io \"authorino-leader-election-role\" not found"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	/home/eguzki/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:329
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/home/eguzki/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:266
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
	/home/eguzki/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:227
```

- [x] Refactor serviceaccount reconciler
- [x] ClusterRoleBinding for `authorino-manager-role` cluster role
- [x] RoleBinding for `authorino-manager-role` cluster role
- [x] ClusterRoleBinding reconciler for `authorino-manager-k8s-auth-role` cluster role
- [x] Role  `authorino-leader-election-role` reconciler
- [x] RoleBinding reconciler for `authorino-leader-election-role` role
- [X] (cluster)rolebinding mutators more flexible. Allow existing subjects and ensure desired ones.

Additionally:
- [X] Add authorino CR finalizer for cleanup logic
- [X] embed logger into context
- [X] makefile targets to run kind local cluster

### Verifications steps

* Checkout this branch

* Run cluster
```
make kind-create-cluster
```

* Install CRDs and other resources (cluster roles, ...)
```
make install
```

* Run operator locally 

```
make run
```

* Create authorino object.
```yaml
kubectl apply -f - <<EOF
kind: Authorino
metadata:
  name: authorino
spec:
  clusterWide: true
  listener:
    tls:
      enabled: false
  oidcServer:
    tls:
      enabled: false
EOF
```

* Wait for authorino instance to be ready
```shell
kubectl wait --timeout=300s --for=condition=Ready authorino authorino
```

* Verify ClusterRoleBinding for the manager cluster role is created and has required subjects
```shell
kubectl get clusterrolebinding authorino -o yaml
```
The expected result is:
```yaml
kind: ClusterRoleBinding
metadata:
  creationTimestamp: "2025-04-15T20:33:17Z"
  name: authorino
  resourceVersion: "1145"
  uid: 72cc8222-f2dc-4bec-8a91-9be35f61f2c9
roleRef:
  kind: ClusterRole
  name: authorino-manager-role
subjects:
- kind: ServiceAccount
  name: authorino-authorino
  namespace: default
```

* Remove authorino CR
```shell
kubectl delete authorino authorino
```

* Verify ClusterRoleBinding for the manager cluster role no longer exists
```shell
kubectl get clusterrolebinding authorino -o yaml
```
The expected result is:
```yaml
Error from server (NotFound): clusterrolebindings.rbac.authorization.k8s.io "authorino" not found
```

* clean up
```
make kind-delete-cluster
```


pkg/reconcilers/deployment_mutator.go
```diff
@@ -88,17 +93,63 @@ func DeploymentImageMutator(desired, existing *k8sapps.Deployment) bool {
 	return update
 }
 
-func DeploymentVolumesMutator(desired, existing *k8sapps.Deployment) bool {
+func DeploymentImagePullPolicyMutator(desired, existing *k8sapps.Deployment) bool {
 	update := false
 
-	if !reflect.DeepEqual(existing.Spec.Template.Spec.Volumes, desired.Spec.Template.Spec.Volumes) {
-		existing.Spec.Template.Spec.Volumes = desired.Spec.Template.Spec.Volumes
+	if existing.Spec.Template.Spec.Containers[0].ImagePullPolicy != desired.Spec.Template.Spec.Containers[0].ImagePullPolicy {
```
Patryk-Stefanski, line 99: Should we nil check containers here to ensure we don't panic ?

pkg/reconcilers/deployment_mutator.go
```diff
@@ -88,17 +93,63 @@ func DeploymentImageMutator(desired, existing *k8sapps.Deployment) bool {
...
+		update = true
+	}
+
+	return update
+}
+
+func DeploymentContainerArgsMutator(desired, existing *k8sapps.Deployment) bool {
+	update := false
+
+	existingArgs := existing.Spec.Template.Spec.Containers[0].DeepCopy().Args
```
Patryk-Stefanski, line 110: Should we nil check containers here to ensure we don't panic ?

pkg/reconcilers/role_binding_mutator.go
```diff
@@ -2,8 +2,8 @@ package reconcilers
 
 import (
 	"fmt"
-	"reflect"
 
+	authorinoResources "github.com/kuadrant/authorino-operator/pkg/resources"
```
Patryk-Stefanski, line 6: nit: move from 3rd party to authorino imports

pkg/reconcilers/deployment_mutator.go
```diff
@@ -88,17 +93,63 @@ func DeploymentImageMutator(desired, existing *k8sapps.Deployment) bool {
...
+	sort.Slice(existingVolumes, func(i, j int) bool {
+		return existingVolumes[i].Name < existingVolumes[j].Name
+	})
+
+	sort.Slice(desiredVolumes, func(i, j int) bool {
+		return desiredVolumes[i].Name < desiredVolumes[j].Name
+	})
+
+	for i, desiredVolume := range desiredVolumes {
+		if existingVolumes[i].Name != desiredVolume.Name { // comparing only the names has limitation, but more reliable than using reflect.DeepEqual or comparing the marshalled version of the resources
```
Patryk-Stefanski, line 144: Hmm not sure how I feel about this. Is there a better way of documenting this limitation or are we better off also iterating through the sources to update when sources change aswell ?

pkg/reconcilers/authorino_reconciler.go
```diff
@@ -130,131 +139,173 @@ func (r *AuthorinoReconciler) ReconcileAuthorinoServices(ctx context.Context, lo
...
+		Rules:      authorinoResources.GetLeaderElectionRules(),
+	}
+
+	if err := ctrl.SetControllerReference(authorinoInstance, role, r.Scheme); err != nil {
+		return err
+	}
+
+	crud, _, err := r.reconcileResource(ctx, &k8srbac.Role{}, role, CreateOnlyMutator)
+	if err != nil {
+		if crud == "read" {
```
guicassolato, line 254: Something in how we're controlling the flow based on a string returned by the method smells.

Ideally, we could leverage proper typing here instead, by returning an interface that abstracts the details of building the error according to each operation.

I understand this would involve a bigger change, so maybe a couple alternatives to mitigate and reduce the risks could be:
- returning the error already wrapped
- using a constant to identify the operation

pkg/reconcilers/deployment_mutator.go
eguzki: well, yes and no. 

Obviously, yes, the code should check. Other mutators do not, but that's not a really good reason to not do it. On the other hand, the `DeploymentContainerListMutator` comes first and makes sure that there number of containers is the same, and the desired deployment has at least one container.. so,... adding those checks but not really needed IMO

pkg/reconcilers/authorino_reconciler.go
eguzki: I was thinking about that as well. I agree it would be much better than string. 

Error type?

I was also looking at the object being returned by the `reconcileResource`.. it might not be necessary. But for follow up work

controllers/authorino_controller.go
```diff
@@ -19,17 +19,15 @@ package controllers
...
 
 	k8sapps "k8s.io/api/apps/v1"
 	k8score "k8s.io/api/core/v1"
-	k8srbac "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
+	"github.com/go-logr/logr"
```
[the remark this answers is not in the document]
eguzki: done

pkg/reconcilers/deployment_mutator.go
eguzki: done

pkg/reconcilers/deployment_mutator.go
eguzki: done

pkg/reconcilers/role_binding_mutator.go
eguzki: done

Patryk-Stefanski (approved):

Issue #262, opened 2025-05-27
Duplicate VolumeMount Path Causes Authorino Deployment Creation Failure


KevFan:
**Describe the bug**
When applying an `Authorino` custom resource with listener tls enable and oidc server tls enabled, the `Authorino Operator` fails to reconcile the corresponding `Deployment`. The error occurs because the generated `Deployment` contains duplicate volumeMounts with the same mountPath value (`/etc/ssl/certs/tls.crt`) in the first container. Kubernetes requires each mountPath to be unique per container, causing the `Deployment` to be rejected and reconciliation to fail.

```log
{"level":"info","ts":"2025-05-27T13:44:19Z","logger":"authorino-operator.controller.Authorino","msg":"create object","authorino":{"name":"authorino","namespace":"default"},"kind":"v1.Deployment","name":"authorino","namespace":"default"}
{"level":"error","ts":"2025-05-27T13:44:19Z","logger":"authorino-operator.controller.Authorino","msg":"Reconciler error","controller":"authorino","controllerGroup":"operator.authorino.kuadrant.io","controllerKind":"Authorino","Authorino":{"name":"authorino","namespace":"default"},"namespace":"default","name":"authorino","reconcileID":"f71295bb-8de5-4957-ac9e-953a832f47e7","error":"failed to reconcile authorino Deployment resource, err: failed to create authorino Deployment resource, err: Deployment.apps \"authorino\" is invalid: spec.template.spec.containers[0].volumeMounts[2].mountPath: Invalid value: \"/etc/ssl/certs/tls.crt\": must be unique","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:329\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:266\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:227"}
```

[...]

PR #263, opened 2025-05-29
fix: use DefaultOidcTlsCertPath

KevFan:
# Description
Closes: https://github.com/Kuadrant/authorino-operator/issues/262

- Fixes deployment creation error when both listener and oidc tls is enabled. This was caused by `DefaultTlsCertPath` being instead of `DefaultOidcTlsCertPath` for oidc causing a mount path duplication
- Updated and added integration test to deployment is created successully for tls enabled permutations

### DIFF
--- a/controllers/authorino_controller_test.go
+++ b/controllers/authorino_controller_test.go
@@ -16,6 +16,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/utils/env"
+	"k8s.io/utils/pointer"
 	controllerruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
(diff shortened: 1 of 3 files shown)


controllers/authorino_controller_test.go
```diff
@@ -417,7 +448,10 @@ func newFullAuthorinoInstance() *api.Authorino {
 			EvaluatorCacheSize:       &cacheSize,
 			Listener: api.Listener{
 				Tls: api.Tls{
-					Enabled: &tlsDisabled,
+					Enabled: &tslEnable,
+					CertSecret: &k8score.LocalObjectReference{
+						Name: testCertSecretName,
+					},
```
KevFan, line 454: With the default tested authorino instance now having tls enabled at listener and oidc server, the deployment for this configuration is now tested by the exisitng test at:
https://github.com/Kuadrant/authorino-operator/blob/db4674db97df4a9fe6eb8a835098357b63524580/controllers/authorino_controller_test.go#L123-L162

And the container args for the oidc server tls cert and key path is tested at:
https://github.com/Kuadrant/authorino-operator/blob/db4674db97df4a9fe6eb8a835098357b63524580/controllers/authorino_controller_test.go#L484-L486

Patryk-Stefanski (approved):
Verification steps were successful and lgtm

PR #282, opened 2025-10-27
Avoid using `/etc/ssl/certs` for TLS files

jhernand:
Currently when Authorino is configured for TLS the required key and certificate files (`tls.key` and `tls.crt`, for example) are mounted into the `/etc/ssl/certs` directory. This makes it difficult to mount custom CA certificates in that same directory, which is the place where Go will read them. For example, if the `ca-bundle` configmap contains a set of trusted CA certificates, a configuration like this should make Authorino trust them:

```
volumes:
  items:
  - name: ca-bundle
    configMaps:
    - ca-bundle
    mountPath: /etc/ssl/certs
```

But instead it results in the following error when trying to start the pod:

```
failed to create containerd task:
failed to create shim task:
OCI runtime create failed:
runc create failed:
unable to start container process: error during container init:
error mounting "/var/lib/kubelet/pods/2b793047-e33e-4cca-b6c3-3d8ced30cb77/volume-subpaths/tls-cert/authorino/1" to rootfs at "/etc/ssl/certs/tls.crt":
mount src=/var/lib/kubelet/pods/2b793047-e33e-4cca-b6c3-3d8ced30cb77/volume-subpaths/tls-cert/authorino/1, dst=/etc/ssl/certs/tls.crt, dstFd=/proc/thread-self/fd/8, flags=MS_RDONLY|MS_BIND|MS_REC:
not a directory
```

That happens when trying in a Kind cluster, may not happen in other Kubernetes flavours.

A possible way to avoid this is to mount the `tls.key` and `tls.crt` files in a different directory, for example `/etc/authorino`. That is what this patch changes.

### DIFF
--- a/pkg/reconcilers/consts.go
+++ b/pkg/reconcilers/consts.go
@@ -59,10 +59,11 @@ const (
 	FlagMaxHttpRequestBodySize         string = "max-http-request-body-size"
 
 	// defaults
-	DefaultTlsCertPath         string = "/etc/ssl/certs/tls.crt"
-	DefaultTlsCertKeyPath      string = "/etc/ssl/private/tls.key"
-	DefaultOidcTlsCertPath     string = "/etc/ssl/certs/oidc.crt"
-	DefaultOidcTlsCertKeyPath  string = "/etc/ssl/private/oidc.key"
+	DefaultConfigPath          string = "/etc/authorino"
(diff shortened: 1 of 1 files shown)


jhernand:
@guicassolato a friendly remider.

PR #303, opened 2026-04-01
ci: adjust nightlies cron to build earlier

averevki:
With the change to [summer time](https://en.wikipedia.org/wiki/Daylight_saving_time) a couple of days ago, and also, probably, because of the downtime issues happening with github currently, nightly kuadrant builds are now getting to the quay registry as late as 5:30 AM. With QE also having nightly pipelines and tasks that utilize nightly kuadrant image being built, these pipelines are mistimed now, and are being executed before image manages to get uploaded to quay.io. 

With this PR, I'm making nightlies build a bit earlier in the night, so we can have a working build as earlier as possible for any other existing and future nightly tasks.

## Summary by CodeRabbit

* **Chores**
  * Updated the nightly build process scheduling to optimise infrastructure operations.

### DIFF
--- a/.github/workflows/build-images-nightly.yaml
+++ b/.github/workflows/build-images-nightly.yaml
@@ -2,7 +2,7 @@ name: Schedule nightly build
 
 on:
   schedule:
-    - cron: "0 0 * * *"
+    - cron: "0 23 * * *"
   workflow_dispatch: {}
 
 jobs:


KevFan (approved):

eguzki (commented):
Heads up. 

This is 23:00.. are we ok with that?

KevFan:
> Heads up.
> 
> This is 23:00.. are we ok with that?

That's good with me anyway

eguzki:
So, limitador will build on Day D and kuadrant operator on Day D+1. Kuadrant operator will try to find reference of the authorino nightly build on Day D+1.

averevki:
@eguzki yeah sorry I was changing this with UTC+1/2 zones in my mind. And now I've noticed that it's probably true that nightly images would have the different day tag when are built at 23:00 and then at 0:30, using UTC+0 also for `date` commands in workflows.

I will look into different solutions, but now I'm actually not sure if we can configure nightly jobs around the summer time in Europe so all of them are executed strictly at night, throughout the entire year. Moving the workflows which use nightly images later in the day seems like the optimal option now to me

averevki:
Closing in favor of https://github.com/Kuadrant/kuadrant-operator/pull/1858. It was decided to only adjust the kuadrant-operator nightly building time, seems like dependencies are built the earliest possible time already (00:00)

PR #306, opened 2026-05-12
Add quay image expiry to nightly and branch workflows

averevki:
Add quay image expiry parameter to the base image building workflow, and set following image expiries on workflows: 
| Image Type | Expiry |
  |---|---|
  | Nightly images | 2w |
  | All other images | never |

This setup was mirrored from https://github.com/Kuadrant/limitador-operator/

## Release Notes

* **Chores**
  * Configured container image expiration on the registry. Main releases persist indefinitely; nightly builds automatically expire after two weeks, reducing unnecessary storage.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/authorino-operator/pull/306)

### DIFF
--- a/.github/workflows/build-images-branches.yaml
+++ b/.github/workflows/build-images-branches.yaml
@@ -17,4 +17,3 @@ jobs:
       operatorTag: ${{ github.ref_name }}
       authorinoVersion: latest
       relatedImageAuthorino: quay.io/kuadrant/authorino:latest
-      quayImageExpiry: 1w
--- a/.github/workflows/build-images-branches.yaml
+++ b/.github/workflows/build-images-branches.yaml
@@ -17,3 +17,4 @@ jobs:
       operatorTag: ${{ github.ref_name }}
       authorinoVersion: latest
       relatedImageAuthorino: quay.io/kuadrant/authorino:latest
+      quayImageExpiry: 1w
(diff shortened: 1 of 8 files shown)


.github/workflows/build-images-branches.yaml
```diff
@@ -17,3 +17,4 @@ jobs:
       operatorTag: ${{ github.ref_name }}
       authorinoVersion: latest
       relatedImageAuthorino: quay.io/kuadrant/authorino:latest
+      quayImageExpiry: 1w
```
guicassolato: I'd rather not expire branch images. They may be referenced in demos and tutorials.

I think some manual cleanup from time to time should be all right.

guicassolato (changes requested):

.github/workflows/build-images-branches.yaml
averevki: Makes sense, removed this

Makefile
```diff
@@ -326,6 +326,7 @@ bundle: manifests kustomize operator-sdk yq ## Generate bundle manifests and met
 	# Roll back edit
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${DEFAULT_OPERATOR_IMAGE}
 	$(MAKE) bundle-custom-modifications
+	echo "$$QUAY_EXPIRY_TIME_LABEL" >> bundle.Dockerfile
```
guicassolato: I wonder if this should be part of `bundle-custom-modifications`. WDYT?

averevki: yeah this will be a better place, I moved it

guicassolato (approved):

PR #308, opened 2026-05-15
Adopt RFC 0018 release branch naming (release-X.Y)

thomasmaas:
## Summary

Updates the release workflow to use minor-level branch naming without `v` prefix per [RFC 0018](https://github.com/Kuadrant/architecture/blob/main/rfcs/0018-release-branch-naming.md).

## Changes

**Granularity change**: the workflow previously created a new branch per patch release (`release-v0.24.0`, `release-v0.24.1`). Now it uses one branch per minor (`release-0.24`) — patch releases are new tags on the same branch.

- Add a `Derive release branch name` step that computes `release-X.Y` from the version input
- Change `startsWith` check from `release-v` to `release-`
- Use derived branch name for `git push` and `target_commitish`
- `RELEASE.md:82` left as-is (historical link to `release-v0.10.0` which is a real existing branch)

Part of Kuadrant/kuadrant-operator#1981

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -30,7 +30,7 @@ jobs:
       - name: Validate version input
         run: |
           version="${{ github.event.inputs.operatorVersion }}"
-          if ! echo "$version" | grep -qE '^[0-9]+\.[0-9]+\.[0-9]+([+-].*)?$'; then
+          if ! echo "$version" | grep -qE '^[0-9]+\.[0-9]+\.[0-9]+([+-].+)?$'; then
             echo "Error: version must be semver format (e.g., 0.24.1, 0.25.0-rc1)"
             exit 1
           fi
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -40,10 +40,15 @@ jobs:
         with:
           go-version-file: go.mod
         id: go
-      - name: Create release branch
-        if: ${{ !startsWith(github.event.inputs.gitRef, 'release-v') }}
+      - name: Derive release branch name
+        id: release-branch
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -40,10 +40,22 @@ jobs:
...
           go-version-file: go.mod
         id: go
+      - name: Derive release branch name
+        id: release-branch
+        run: |
+          version="${{ github.event.inputs.operatorVersion }}"
+          if ! echo "$version" | grep -qE '^[0-9]+\.[0-9]+\.[0-9]+([+-].*)?$'; then
+            echo "Error: version must be semver format (e.g., 0.24.1, 0.25.0-rc1)"
+            exit 1
+          fi
```
guicassolato: If the goal is to validate the user input – even if it's so this particular step doesn't break –, should we do it in a separate step, early on in the job, before starting checking out the code and installing stuff?

guicassolato: In time: for a proper validation, we probably want:


```suggestion
          if ! echo "$version" | grep -qE '^[0-9]+\.[0-9]+\.[0-9]+([+-].+)?$'; then
            echo "Error: version must be semver format (e.g., 0.24.1, 0.25.0-rc1)"
            exit 1
          fi
```

guicassolato (dismissed):
I left a comment about validating user input, but generally this looks good. Let's give it a try to make sure it all works as expected, including the chained action for releasing the Helm charts.

.github/workflows/release.yaml
thomasmaas: yes, fixed that edge case. Thanks.

guicassolato (approved):

PR #313, opened 2026-05-27
feat: add USE_IMAGE_DIGESTS flag for digest-based image references

mikenairn:
## Summary

Adds optional USE_IMAGE_DIGESTS flag to enable digest-based image references in operator bundles.

## Changes

- Added `USE_IMAGE_DIGESTS` flag (default: false) to enable SHA256 digest references
- Created `BUNDLE_GEN_FLAGS` variable with conditional `--use-image-digests`
- Added bundle post-generation step to deduplicate relatedImages (operator-sdk creates duplicates when using digests)
- Fixed bundle validation to run after all bundle modifications complete

Digest-based references enable ImageDigestMirrorSet to redirect image pulls to internal registries.

## Related

Related to https://github.com/Kuadrant/kuadrant-operator/issues/1894

### DIFF
--- a/make/catalog.mk
+++ b/make/catalog.mk
@@ -57,7 +57,7 @@ catalog-build: ## Build a catalog image.
 # Push the catalog image.
 .PHONY: catalog-push
 catalog-push: ## Push a catalog image.
-	$(MAKE) docker-push IMG=$(CATALOG_IMG)
+	$(MAKE) docker-push OPERATOR_IMAGE=$(CATALOG_IMG)
 
 deploy-catalog: kustomize yq ## Deploy operator to the K8s cluster specified in ~/.kube/config using OLM catalog image.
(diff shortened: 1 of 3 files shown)


make/catalog.mk
```diff
@@ -57,7 +57,7 @@ catalog-build: ## Build a catalog image.
 # Push the catalog image.
 .PHONY: catalog-push
 catalog-push: ## Push a catalog image.
-	$(MAKE) docker-push IMG=$(CATALOG_IMG)
+	$(MAKE) docker-push OPERATOR_IMAGE=$(CATALOG_IMG)
```
guicassolato, line 60: I guess this target hasn't been much used lately. Thanks for the fix anyway.

guicassolato (approved):
I've tried on `main` and on `release-0.25` and it seems to work as expected.

make/catalog.mk
mikenairn: Probably never used by anyone. tbh, I'm starting to think a lot of what we have in these repos could just be removed to make our life easier. Do we even need this catalog?

PR #325, opened 2026-06-20
Use build.yaml for release configuration in CI workflows

guicassolato:
## Problem
The `build-images-branches.yaml` workflow currently passes hardcoded values (`VERSION=0.0.0`, `AUTHORINO_VERSION=latest`) to the base workflow. This causes issues for branches targeting release branches (e.g., backport branches for patch releases) because:

1. The hardcoded versions don't match what's committed in the branch
2. `make verify-manifests` and `make verify-bundle` fail in CI because manifests are generated with different versions than what's committed
3. Release branches need to use specific versions, not `0.0.0` and `latest`

## Solution
Enable the Makefile and CI workflows to read release configuration from `build.yaml` when present:

- **For dev branches** (no `build.yaml`): Falls back to defaults (`0.0.0`, `latest`) - no change in behavior
- **For release branches** (has `build.yaml`): Automatically uses the correct versions from `build.yaml`

### Makefile
- Reads `VERSION`, `AUTHORINO_VERSION`, and `CHANNELS` from `build.yaml` if file exists
- Uses grep/sed for parsing (no yq dependency needed at variable definition time)
- Includes whitespace trimming for robustness
- Falls back to git SHA and 'latest' when `build.yaml` doesn't exist

### GitHub Workflow (build-images-branches.yaml)
- Adds `detect-config` job to extract configuration from `build.yaml`
- Passes extracted values to `build-images-base` workflow
- Falls back to `0.0.0`/`latest` when `build.yaml` not found

## Alternative Considered
Commit `build.yaml` in main with default values (`VERSION=0.0.0`, `AUTHORINO_VERSION=latest`, etc.) and simplify the Makefile to always read from the build file without fallback logic.

This alternative was **discarded for now** to make the change less disruptive - `build.yaml` continues to only exist in release branches. However, we're open to revisiting this decision in the future if having a committed `build.yaml` in main proves beneficial.

## Benefits
- ✅ Release/backport branches can use `build.yaml` to set correct versions automatically
- ✅ CI verification passes because generated manifests match committed ones
- ✅ No change for dev branches (maintain current behavior)
- ✅ Cleaner than environment-specific workarounds

## Testing
- ✅ Tested on release backport branches (v0.23.3, v0.24.1, v0.25.1)
- ✅ Variable extraction works correctly with whitespace trimming
- ✅ Falls back gracefully when `build.yaml` doesn't exist

---
Developed while preparing patch releases v0.23.3, v0.24.1, and v0.25.1

Contains commits of #326, which should be merged first for a clean history.

## Release Notes

* **Chores**
  * Improved build infrastructure by detecting build parameters from `build.yaml` when present, including versioning and release channel defaults, with sensible fallbacks when the file is missing.
  * Updated branch image builds to use these dynamically detected values and tightened registry credential passing.
  * Enhanced the Makefile to derive release-related variables from `build.yaml`, defaulting channels to `stable` when unset.

### DIFF
--- a/.github/workflows/build-images-branches.yaml
+++ b/.github/workflows/build-images-branches.yaml
@@ -8,12 +8,53 @@ on:
   workflow_dispatch: {}
 
 jobs:
+  detect-config:
+    name: Detect build configuration
+    runs-on: ubuntu-latest
+    outputs:
+      operatorVersion: ${{ steps.config.outputs.operatorVersion }}
+      authorinoVersion: ${{ steps.config.outputs.authorinoVersion }}
+      authorinoImage: ${{ steps.config.outputs.authorinoImage }}
+      channels: ${{ steps.config.outputs.channels }}
+      defaultChannel: ${{ steps.config.outputs.defaultChannel }}
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v4
+      - name: Read configuration from build.yaml
+        id: config
+        run: |
+          BUILD_FILE="build.yaml"
+          if [ -f "$BUILD_FILE" ]; then
+            echo "build.yaml found, extracting configuration"
+            # Read values using grep/sed (same approach as Makefile, with trimming)
+            OPERATOR_VERSION=$(grep -E '^\s+version:' "$BUILD_FILE" 2>/dev/null | sed -E 's/.*version:\s*//;s/^[[:space:]]+//;s/[[:space:]]+$//' || echo "0.0.0")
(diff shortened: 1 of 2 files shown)


Boomatang:
I am going to say hold off on this for a little bit. I like the idea, and there is currently a RFC looking to standardize this across the different repo. It is planned to be talked about in the community call on June 23rd. 

Here is a link to the RFC https://github.com/Kuadrant/limitador/pull/494 , and here is a link to a draft PR I have in autorino, https://github.com/Kuadrant/authorino/pull/634, it still needs a bit of testing, but the layout would be the same.

eguzki:
This is partially covering https://github.com/Kuadrant/authorino-operator/issues/329.

PR #328, opened 2026-06-22
Authorino manifests sync automatic worflow

averevki:
Adds a GitHub Actions workflow that automatically regenerates the AuthConfig CRD and related manifests when triggered by upstream changes in the [authorino](https://github.com/Kuadrant/authorino) repo.

The workflow runs `make manifests bundle helm-build` and opens a PR with the updated files. It maintains a single rolling PR (`sync/authorino-manifests` branch), updating it on each trigger. Manifest updates with only a `createdAt` timestamp change are skipped.

### Triggered by:
  - `repository_dispatch` from authorino [sync-operator-manifests](https://github.com/Kuadrant/authorino/blob/sync-manifests-workflow/.github/workflows/sync-operator-manifests.yaml) (https://github.com/Kuadrant/authorino/pull/640) workflow
  - Manual `workflow_dispatch` with optional `authorinoVersion` input

## Summary by CodeRabbit

* **Chores**
  * Added an automated sync process for manifests that can be run manually or triggered by an event.
  * Prevents duplicate runs by cancelling older in-progress syncs.
  * Opens a pull request only when there are meaningful changes, reducing noisy updates.


.github/workflows/sync-authorino-manifests.yaml
```diff
@@ -0,0 +1,69 @@
...
+  group: authorino-manifests-sync
+  cancel-in-progress: true
+
+jobs:
+  sync-manifests:
+    name: Sync manifests and create PR
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
```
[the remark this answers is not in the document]
averevki: Similar Kuadrant org workflows also use tags instead of the hashes on github actions. This is a good point, but we would need to also update other actions to also use the hashes instead. It should be done in another PR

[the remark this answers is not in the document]
averevki: Can you create this issue in the Kuadrant/kuadrant-operator github project?

.github/workflows/sync-authorino-manifests.yaml
```diff
@@ -0,0 +1,72 @@
...
+          base: main
+          branch: sync/authorino-manifests
+          delete-branch: true
+          title: Update Authorino manifests
+          body: |
+            Automated sync of Authorino AuthConfig CRD and related manifests.
+
+            Synced by running `make manifests bundle helm-build AUTHORINO_VERSION=${{ github.event.inputs.authorinoVersion || 'latest' }}`.
+
+            [Recent Authorino manifest changes](https://github.com/Kuadrant/authorino/commits/main/install)
```
guicassolato, line 62: Wondering if we should use `github.event.inputs.authorinoVersion` here and avoid generating manifests for one version while pointing the link always to `main`.

Basically:
- if `github.event.inputs.authorinoVersion == 'latest'` → `git_ref = main`
- else → `git_ref = "v${{ github.event.inputs.authorinoVersion }}"`

WDYT?

averevki: It makes sense to change. However, I'm thinking now, if we should instead hard-code this action to work only with main branches instead. You think there is a case when merging manifests from an another authorino branch to the authorino-operator main could be useful?

guicassolato: Other Authorino versions usually make sense in the release branches, where it's expected to be driven by the `build.yaml` file. IOW, only `main` branch is fine.

averevki: It can only sync main branches now :+1:

guicassolato (approved):

PR #342, opened 2026-07-22
fix: corrected build argument for docker

DaliborD45:
The ACTUAL_DEFAULT_AUTHORINO_IMAGE argument was being passed to the Docker build command, but this did not match the argument expected by the Dockerfile. The Dockerfile expects the DEFAULT_AUTHORINO_IMAGE build argument, so this change aligns the build command with the Dockerfile configuration.


guicassolato (approved):
Thanks @DaliborD45!

PR #349, opened 2026-08-14
fix: namespace-qualify cluster-scoped ClusterRoleBinding names

### LINKED ISSUE #348: ClusterRoleBinding name collision between same-named Authorino instances in different namespaces
**Describe the bug**

  When two cluster scoped Authorino instances share the same CR name (e.g. both named `authorino`) but live in different namespaces, they collide on a single cluster-scoped `ClusterRoleBinding` and overwrite each other's permissions.

  The `ClusterRoleBinding` name for Authorino's Kubernetes `TokenReview`/`SubjectAccessReview` role is derived from the CR name only:

  `<crName>-authorino-k8s-auth`

  Because `ClusterRoleBindings` are cluster-scoped and the name does not include the namespace, two instances both named `authorino` resolve to the same binding, `authorino-authorino-k8s-auth`. The operator reconciles this binding with Server-Side Apply, setting subjects to just its own ServiceAccount.
  Since subjects is an atomic list under SSA, each reconciliation replaces the field rather than merging — so whichever instance reconciles last wins, and the other instance's ServiceAccount is evicted from the binding.

  The affected instance then fails Kubernetes token reviews, e.g.:
```
  tokenreviews.authentication.k8s.io is forbidden:
    User "system:serviceaccount:kuadrant-system:authorino-authorino"
    cannot create resource "tokenreviews" in API group
    "authentication.k8s.io" at the cluster scope
```
  This is a follow-up to #291 / PR #292. PR #292 made binding names instance-specific by CR name, which fixes instances with different names, but it does not disambiguate by namespace — so same-named instances in different namespaces still collide.

  **To Reproduce**

  1. Deploy a cluster-wide Authorino instance named authorino in namespace `ns-a`.
  2. Deploy a second cluster-wide Authorino instance, also named authorino, in namespace `ns-b`.
  3. Inspect the shared ClusterRoleBinding: 
  ```
oc get clusterrolebinding authorino-authorino-k8s-auth -o jsonpath='{.subjects[0].namespace}'
  ```
  3. It reports only one namespace — whichever instance reconciled last.
  4. Trigger a re-reconcile of the other instance (e.g. annotate it), then confirm the first instance's ServiceAccount can no longer create
  tokenreviews:
  ```
   oc auth can-i create tokenreviews.authentication.k8s.io \
    --as=system:serviceaccount:ns-a:authorino-authorino
  # -> no
  ```
  **Expected behavior**

  Each Authorino instance should get its own ClusterRoleBinding that is not affected by any other instance, regardless of CR name or namespace. Both instances should retain the ability to create tokenreviews/subjectaccessreviews, and inference/auth requests routed through either instance should succeed.

  **Screenshots**

  N/A

  **Additional context**

  - Discovered in an RHOAI 3.4 + Red Hat Connectivity Link (RHCL) LLM-D setup, where Kuadrant creates an Authorino instance named `authorino` in kuadrant-system alongside a pre-existing authorino instance in another namespace. All authenticated inference requests to the gateway returned HTTP
  `401`, and the associated AuthPolicy reported `Enforced: False`.
  - The `authorino-manager-k8s-auth-role` `ClusterRole` itself is correct; only the binding's subjects field is affected.
  - Proposed fix: include the namespace in the cluster-scoped binding name (e.g. `<namespace>-<crName>-authorino-k8s-auth`) so bindings are globally
  unique per instance, plus a best-effort cleanup of the legacy `<crName>-<suffix>` names on upgrade.
  - Workaround: manually create a uniquely-named `ClusterRoleBinding` for the affected instance:
  ```
  kind: ClusterRoleBinding
  metadata:
    name: kuadrant-system-authorino-k8s-auth-manual
  roleRef:
    kind: ClusterRole
    name: authorino-manager-k8s-auth-role
  subjects:
  - kind: ServiceAccount
    name: authorino-authorino
    namespace: kuadrant-system
  ```

KevFan:
ClusterRoleBindings are not namespaced, but their names were derived from the CR name only (<crName>-<suffix>). Two Authorino instances sharing a CR name in different namespaces collided on a single binding and evicted each other's ServiceAccount under Server-Side Apply, breaking Kubernetes TokenReview/SubjectAccessReview and returning HTTP 401.

Include the namespace in the binding name (<namespace>-<crName>-<suffix>) and best-effort delete the legacy-named bindings on reconcile. Add regression tests covering same-named instances across namespaces and the legacy binding migration.

Fixes #348 https://redhat.atlassian.net/browse/RHOAIENG-71904

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented permission conflicts when Authorino instances share the same name across different namespaces.
  * Ensured each instance receives and retains its correct Kubernetes authentication binding.
  * Automatically removes legacy bindings and creates namespace-specific replacements during reconciliation.
* **Tests**
  * Added regression and migration coverage for namespace-specific bindings and legacy cleanup.

### DIFF
--- a/controllers/authorino_controller.go
+++ b/controllers/authorino_controller.go
@@ -151,12 +151,12 @@ func (r *AuthorinoReconciler) cleanupClusterScopedPermissions(ctx context.Contex
 	// namespaced ones are garbage collected automatically by k8s because of the owner reference
 
 	// Delete instance-specific ClusterRoleBindings
-	managerBinding := authorinoResources.GetAuthorinoClusterRoleBinding(crName, reconcilers.AuthorinoManagerClusterRoleBindingName, reconcilers.AuthorinoManagerClusterRoleName, sa, labels)
+	managerBinding := authorinoResources.GetAuthorinoClusterRoleBinding(crNamespacedName.Namespace, crName, reconcilers.AuthorinoManagerClusterRoleBindingName, reconcilers.AuthorinoManagerClusterRoleName, sa, labels)
 	if err := r.Client.Delete(ctx, managerBinding); err != nil && !errors.IsNotFound(err) {
 		r.Log.Error(err, "failed to delete ClusterRoleBinding", "name", managerBinding.Name)
 	}
 
-	k8sAuthBinding := authorinoResources.GetAuthorinoClusterRoleBinding(crName, reconcilers.AuthorinoK8sAuthClusterRoleBindingName, reconcilers.AuthorinoK8sAuthClusterRoleName, sa, labels)
+	k8sAuthBinding := authorinoResources.GetAuthorinoClusterRoleBinding(crNamespacedName.Namespace, crName, reconcilers.AuthorinoK8sAuthClusterRoleBindingName, reconcilers.AuthorinoK8sAuthClusterRoleName, sa, labels)
 	if err := r.Client.Delete(ctx, k8sAuthBinding); err != nil && !errors.IsNotFound(err) {
 		r.Log.Error(err, "failed to delete ClusterRoleBinding", "name", k8sAuthBinding.Name)
 	}
--- a/pkg/resources/k8s_rbac.go
+++ b/pkg/resources/k8s_rbac.go
@@ -56,19 +56,15 @@ func GetSubjectForRoleBinding(serviceAccount *k8score.ServiceAccount) k8srbac.Su
 	}
 }
 
-func subjectIncluded(subjects []k8srbac.Subject, subject k8srbac.Subject) bool {
+// SubjectIncluded reports whether the given subject is present in the subjects slice,
+// matching on Kind, Name and Namespace.
+func SubjectIncluded(subjects []k8srbac.Subject, subject k8srbac.Subject) bool {
 	for _, s := range subjects {
 		if s.Kind == subject.Kind && s.Name == subject.Name && s.Namespace == subject.Namespace {
 			return true
 		}
 	}
 	return false
-}
-
-// SubjectIncluded reports whether the given subject is present in the subjects slice,
-// matching on Kind, Name and Namespace.
-func SubjectIncluded(subjects []k8srbac.Subject, subject k8srbac.Subject) bool {
-	return subjectIncluded(subjects, subject)
 }
 
 func GetLeaderElectionRules() []k8srbac.PolicyRule {
@@ -94,26 +90,4 @@ func GetLeaderElectionRules() []k8srbac.PolicyRule {
 			Verbs:     []string{"get", "list", "create", "update"},
 		},
 	}
-}
-
-// MergeBindingSubject merges desired subject slice into the existing slice.
-//
-// The subject entries included in "existing" slice that are not included in the "desired" slice are preserved.
-//
-// It returns true if the existing slice was modified (i.e., at least one subject was added),
-// and false otherwise.
-func MergeBindingSubject(desired []k8srbac.Subject, existing *[]k8srbac.Subject) bool {
-	if existing == nil {
-		return false
-	}
-
-	update := false
-	for idx := range desired {
-		if !subjectIncluded(*existing, desired[idx]) {
-			*existing = append(*existing, desired[idx])
-			update = true
-		}
-	}
-
-	return update
 }
--- a/pkg/resources/k8s_rbac.go
+++ b/pkg/resources/k8s_rbac.go
@@ -65,6 +65,12 @@ func subjectIncluded(subjects []k8srbac.Subject, subject k8srbac.Subject) bool {
 	return false
 }
 
+// SubjectIncluded reports whether the given subject is present in the subjects slice,
+// matching on Kind, Name and Namespace.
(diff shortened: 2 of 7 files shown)


KevFan:
@coderabbitai review

KevFan:
@coderabbitai review

KevFan:
@coderabbitai review

pkg/resources/k8s_util.go
```diff
@@ -25,6 +27,50 @@ func authorinoRoleBindingName(crName, roleBindingNameSuffix string) string {
...
+	if available < 0 {
+		available = 0
+	}
+
+	truncatedName := crName
+	if len(truncatedName) > available {
+		truncatedName = truncatedName[:available]
+	}
+
+	return fmt.Sprintf("%s.%s%s-%s", namespace, truncatedName, hash, clusterRoleBindingNameSuffix)
```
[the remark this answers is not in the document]
KevFan: Thanks for flagging this. I looked into it, but I'm going to decline the change as proposed — a few things don't line up:

  **Wrong location.** The comment points at `pkg/resources/k8s_util.go lines 47–75`, but that's `authorinoClusterRoleBindingName`, a pure string builder. There's no `reconcileResource`, `ApplyResource`, or `ForceOwnership` in that file — those live in `pkg/reconcilers/authorino_reconciler.go`.

  **The underlying concern is real but very narrow.** There is one non-injectivity: since a CR name is a DNS-1123 subdomain and may contain dots, a legacy binding (`<crName>-<suffix>`) for a CR named e.g. `team-a.gateway` is byte-identical to the new canonical name for namespace `team-a` + CR `gateway`. Worth noting the boundaries:
  - Among all instances on the new scheme the naming is injective — a namespace can't contain `.`, so the prefix before the first . is unambiguously the namespace. Two new instances can't collide. 
  - The collision is a migration-window artifact only, and only when a legacy CR name reconstructs another instance's `namespace.crName`.

  It's already mostly mitigated. The legacy cleanup now Gets each binding and only deletes it when the subject is this instance's own ServiceAccount (see the ownership-check commit), so it won't delete another instance's binding. On the apply side, the SSA ForceOwnership resolves to the correct canonical owner of that name and only binds the ClusterRole to that instance's own SA — no privilege it wasn't already entitled to. Actual disruption is limited to a mixed-operator-version window and is self-healing once both reconcile.

  **Why not the suggested fix.** Using a "collision-safe migration name instead of ForceOwnership" would make an instance's canonical binding name non-deterministic, which breaks the predictable-naming design and the cleanup logic that depends on it, and would leave the rightful owner with no binding. That's a permanent design regression to cover a rare, transient, self-healing edge case.

  I'll instead tighten the doc comment on authorinoClusterRoleBindingName to document the legacy-vs-new migration caveat, and add a regression test asserting the ownership-checked cleanup does not delete a colliding binding owned by another instance.

pkg/resources/k8s_rbac.go
```diff
@@ -65,6 +65,12 @@ func subjectIncluded(subjects []k8srbac.Subject, subject k8srbac.Subject) bool {
 	return false
 }
 
+// SubjectIncluded reports whether the given subject is present in the subjects slice,
+// matching on Kind, Name and Namespace.
+func SubjectIncluded(subjects []k8srbac.Subject, subject k8srbac.Subject) bool {
+	return subjectIncluded(subjects, subject)
+}
+
```
guicassolato: `subjectIncluded` is only called on one place. I guess exporting it would be OK instead of defining another func. In fact, it looks like the only place where it's used is from another dead code func which, like `UnboundAuthorinoServiceAccountFromClusterRole`, could also be removed: [`MergeBindingSubject`](https://github.com/Kuadrant/authorino-operator/blob/4dd947dd0e7620e277845aea4ba05658b175a632/pkg/resources/k8s_rbac.go#L99).

controllers/authorino_controller.go
guicassolato, line 146: Nit: the finalizer path (this func) only constructs new-format binding names (`<ns>.<crName>-<suffix>`), so it wouldn't clean up legacy-named bindings if they still existed at CR deletion time.

In practice I guess this is a non-issue. Say one upgrades the operator – the controller restarts and reconciles all existing CRs before any deletion events are processed, so `cleanupLegacyClusterRoleBindings` will have already run by then.

Mentioning it just for awareness.

guicassolato (dismissed):
Nice fix! Thanks @KevFan!

I left one minor nit about removing another piece of dead code – another leftover like `reconcilers.UnboundAuthorinoServiceAccountFromClusterRole` from back in the days when we used a singleton ClusterRoleBinding for all cluster-wide instances instead of dedicated ones.

Approving the PR regardless since it's a non-blocker.

pkg/resources/k8s_rbac.go
KevFan: Makes sense, updated to remove [MergeBindingSubject](https://github.com/Kuadrant/authorino-operator/blob/4dd947dd0e7620e277845aea4ba05658b175a632/pkg/resources/k8s_rbac.go#L99) 👍

guicassolato (approved):

PR #351, opened 2026-08-22
feat: add ValidatingAdmissionPolicies to prevent privilege escalation via Authorino/AuthConfig specs

DaliborD45:
## What

  Adds two cluster-scoped `ValidatingAdmissionPolicy` (+ binding) resources that lock
  down spec fields which can be used to escalate privileges. Both allow-list the
  Kuadrant operator ServiceAccount as the only exception.

## Summary by CodeRabbit

- **New Features**
  - Added configurable response-size limits for several authentication and policy HTTP requests, helping prevent excessively large responses.
  - Added separate permissions for changing Authorino images, enabling cluster-wide mode, and configuring superseding host subsets.

- **Bug Fixes**
  - Updated validation to reject restricted Authorino settings unless the requester has the corresponding permission.
  - Standardised validation failures as forbidden requests with field-specific messages.

- **Chores**
  - Updated generated deployment and Helm manifests to include the revised cluster-scoped admission controls.


config/vap/restrict-all-namespaces.yaml
```diff
@@ -0,0 +1,31 @@
...
+  failurePolicy: Fail
+  matchConstraints:
+    resourceRules:
+      - apiGroups: ["authorino.kuadrant.io"]
+        apiVersions: ["v1beta3"]
+        operations: ["CREATE", "UPDATE"]
+        resources: ["authconfigs"]
+  variables:
+    - name: isKuadrantOperator
+      expression: "request.userInfo.username == 'system:serviceaccount:kuadrant-system:kuadrant-operator-controller-manager'"
```
guicassolato, line 15: Could we use Kubernete's [`authorizer.requestResource`](https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/#validation-expression) CEL binding here and abstract the exception case behind RBAC perhaps, thus avoiding the misdirected coupling with the Kuadrant Operator?

E.g.:

```suggestion
    - name: isExempt
      expression: "authorizer.requestResource().check('set-privileged-fields').allowed()"
```

`'set-privileged-fields'` is a synthetic verb.

Authorino Operator could ship the ClusterRole for convenience:

```yaml
kind: ClusterRole
metadata:
  name: authorino-set-privileged-fields
rules:
  - apiGroups: ["authorino.kuadrant.io"]
    resources: ["authconfigs"]
    verbs: ["set-privileged-fields"]
```

Kuadrant Operator would then ship its own ClusterRoleBinding:

```yaml
kind: ClusterRoleBinding
metadata:
  name: kuadrant-operator-authorino-privileged-fields
roleRef:
  kind: ClusterRole
  name: authorino-set-privileged-fields
subjects:
  - kind: ServiceAccount
    name: kuadrant-operator-controller-manager
    namespace: kuadrant-system
```

The final validation expression at line 21 would become: `!variables.wantsAllNamespaces || variables.wasAllNamespaces || variables.isExempt` of course.

In time – for more fine-grained permissions, we could even define more specific verbs. E.g.:
- `apikey-allnamespaces`
- `x509-allnamespaces`

WDYT?

config/vap/restrict-spec-fields.yaml
```diff
@@ -0,0 +1,34 @@
...
+  failurePolicy: Fail
+  matchConstraints:
+    resourceRules:
+      - apiGroups: ["operator.authorino.kuadrant.io"]
+        apiVersions: ["v1beta1"]
+        operations: ["CREATE", "UPDATE"]
+        resources: ["authorinos"]
+  variables:
+    - name: isKuadrantOperator
+      expression: "request.userInfo.username == 'system:serviceaccount:kuadrant-system:kuadrant-operator-controller-manager'"
```
guicassolato: Similarly to https://github.com/Kuadrant/authorino-operator/pull/351/changes#r3841866165:


```suggestion
    - name: isExempt
      expression: "authorizer.requestResource().check('set-privileged-fields').allowed()"
```

And ClusterRole:

```yaml
kind: ClusterRole
metadata:
  name: authorino-set-custom-image
rules:
  - apiGroups: ["operator.authorino.kuadrant.io"]
    resources: ["authorinos"]
    verbs: ["set-custom-image"]
```

config/vap/restrict-all-namespaces.yaml
```diff
@@ -0,0 +1,31 @@
...
+  variables:
+    - name: isExempt
+      expression: "authorizer.requestResource.check('set-privileged-fields').allowed()"
+    - name: wantsAllNamespaces
+      expression: "has(object.spec.authentication) && object.spec.authentication.exists(k, (has(object.spec.authentication[k].apiKey) && has(object.spec.authentication[k].apiKey.allNamespaces) && object.spec.authentication[k].apiKey.allNamespaces) || (has(object.spec.authentication[k].x509) && has(object.spec.authentication[k].x509.allNamespaces) && object.spec.authentication[k].x509.allNamespaces))"
+    - name: wasAllNamespaces
+      expression: "request.operation == 'UPDATE' && has(oldObject.spec.authentication) && oldObject.spec.authentication.exists(k, (has(oldObject.spec.authentication[k].apiKey) && has(oldObject.spec.authentication[k].apiKey.allNamespaces) && oldObject.spec.authentication[k].apiKey.allNamespaces) || (has(oldObject.spec.authentication[k].x509) && has(oldObject.spec.authentication[k].x509.allNamespaces) && oldObject.spec.authentication[k].x509.allNamespaces))"
+  validations:
+    - expression: "!variables.wantsAllNamespaces || variables.wasAllNamespaces || variables.isExempt"
```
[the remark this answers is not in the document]
guicassolato: The way I'm reading this is:
1. user/SA is granted permission (by binding the user/SA's name to the `authorino-set-privileged-fields` ClusterRole)
2. user/SA sets `allNamespaces` `true` in an AuthConfig in the authorised namespace (occasionally cluster-wide)
3. permission for user/SA is revoked (by deleting the binding to the `authorino-set-privileged-fields` ClusterRole)

After that, all updates to the AuthConfigs requested by the user/SA will overlook `allNamespaces: true`, until the field is set to `false` at least once. IOW: by not touching the field, users can perpetuate the permissions once granted beyond revocation.

Makes me wonder if we need `wasAllNamespaces` at all 🤔

Would a simpler rule `!variables.wantsAllNamespaces || variables.isExempt` work? Of course, to avoid breaking changes, all preexisting resources would have to be migrated or the RABC permissions applied beforehand.

guicassolato (changes requested):
This should go after #352, in a phased approach.

Before upgrading to a version that forces these new safe defaults introduced by this PR, downstream projects that currently depend on Authorino and Authorino Operator (e.g.: [Kuadrant Operator](https://github.com/kuadrant/kuadrant-operator)) should be given a buffer to define the required RBAC, to avoid regression.

Steps to upgrade:
1. **Preparation:** the downstream project specifies Roles and RoleBindings for their authorised SAs granting the required privileges that later will be verified by the VAPs.
2. **Enforcement:** Downstream project upgrades to version of Authorino Operator that ships the enforcement of the new safe defaults.
3. **Swap:** Downstream project specifies new RoleBindings to the standard Roles shipped with the new version of the Operator, making the Roles and RoleBindings specified at (1) redundant – ideally along with (2)
4. **Cleanup:** Downstream project cleans up redundant Roles and RoleBindings defined at (1) – immediately after (3)

config/install/manifests.yaml
```diff
@@ -544,3 +580,56 @@ subjects:
...
+    message: 'spec.supersedingHostSubsets: true can only be set by a subject granted
+      the ''set-superseding-host-subsets'' permission on authorinos'
+    reason: Forbidden
+  variables:
+  - expression: authorizer.requestResource.check('set-cluster-wide').allowed()
+    name: isExemptClusterWide
+  - expression: authorizer.requestResource.check('set-image').allowed()
+    name: isExemptImage
+  - expression: authorizer.requestResource.check('set-superseding-host-subsets').allowed()
+    name: isExemptSupersedingHostSubsets
```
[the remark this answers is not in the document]
DaliborD45: Wildcard-role is intended, so its okay

config/vap/restrict_spec_fields_vap.yaml
```diff
@@ -0,0 +1,43 @@
...
+metadata:
+  name: authorino-restrict-spec-fields
+spec:
+  failurePolicy: Fail
+  matchConstraints:
+    resourceRules:
+      - apiGroups: ["operator.authorino.kuadrant.io"]
+        apiVersions: ["v1beta1"]
+        operations: ["CREATE", "UPDATE"]
+        resources: ["authorinos"]
```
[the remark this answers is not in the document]
DaliborD45: apiKey.allNamespaces and x509.allNamespaces are resolved in other PR, so you dont have to bother with this

DaliborD45:
@guicassolato I have created the feature as we stated in the docs PR #352. I do not know though when it should be merged. Should it be merged only after the release of the docs to the customers, or can it be merged now ?

config/vap/restrict_spec_fields_vap.yaml
```diff
@@ -0,0 +1,43 @@
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingAdmissionPolicy
+metadata:
+  name: authorino-restrict-spec-fields
+spec:
+  failurePolicy: Fail
+  matchConstraints:
+    resourceRules:
+      - apiGroups: ["operator.authorino.kuadrant.io"]
+        apiVersions: ["v1beta1"]
```
philbrookes, line 10: Have you considered using * here to cover future versions? Or are we intending to correct this problem in the code before that would be necessary?

philbrookes (approved):
Quick question, but otherwise LGTM

config/vap/restrict_spec_fields_vap.yaml
```diff
@@ -0,0 +1,43 @@
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingAdmissionPolicy
```
guicassolato, line 2: This VAP is still not included in the manifests nor applied by the controller at runtime, right? So the idea is that we create the ClusterRoles for the user but the user still has to apply the VAP manually (hopefully with no changes that could cause a mismatch with the existing roles)?

jasonmadigan:
just for my own understanding re: the rollout for this one: kuadrant-operator creates every authorino with `clusterWide: true `and its clusterrole only has explicit verbs on `authorinos`, so as soon as someone applies this vap it'd be denied until it holds set-cluster-wide, right? 

is that the only verb it'd need, given it never sets `image` or `supersedingHostSubsets`? and, another q:  is the plan that we raise the kuadrant-operator rbac PR first, then a follow-up here to have the operator apply the policy by default, or do we want the controller to own applying it from the start behind some sort of flag?

guicassolato:
> just for my own understanding re: the rollout for this one: kuadrant-operator creates every authorino with `clusterWide: true `and its clusterrole only has explicit verbs on `authorinos`, so as soon as someone applies this vap it'd be denied until it holds set-cluster-wide, right?
> 
> is that the only verb it'd need, given it never sets `image` or `supersedingHostSubsets`? and, another q: is the plan that we raise the kuadrant-operator rbac PR first, then a follow-up here to have the operator apply the policy by default, or do we want the controller to own applying it from the start behind some sort of flag?

@jasonmadigan, I think your understanding is correct and so is your conclusion. At the very least, a PR to the Kuadrant Operator introducing a RoleBinding for the Operator's SA to the ClusterRole granting `set-image` permission has to go hand-to-hand with this one.

config/vap/restrict_spec_fields_vap.yaml
DaliborD45: yes, the VAP is still not included anywhere
I dont know though whether it is desired behavior or no.
When I am thinking about it we probably dont want to apply some policies to customer right away...

config/vap/restrict_spec_fields_vap.yaml
DaliborD45: > Or are we intending to correct this problem in the code before that would be necessary?

Yes thats correct. There will be a third step of fixing this in the code

DaliborD45:
@jasonmadigan 
> is that the only verb it'd need, given it never sets image or supersedingHostSubsets? 

Why only this verb ? Shouldnt kuadrant-operator set all 3 roles on authorino ?

jasonmadigan:
> Why only this verb ? Shouldnt kuadrant-operator set all 3 roles on authorino ?

 yeah, all three is fine. set-cluster-wide is the only one main actually needs, but released versions still set supersedingHostSubsets so binding all three is the safe call?

DaliborD45:
> set-cluster-wide is the only one main actually needs

Why? Every role exists because it is intended to restrict unauthorized access to a specific resource. Or do you mean that `set-cluster-wide` is the only one main needs because it is the only role currently used by Authorino?

> but released versions still set supersedingHostSubset

I also don't think I fully understand what you mean by this. Does the current Authorino version still set `supersedingHostSubset` somewhere? If so, wouldn't it need the corresponding role as well?

jasonmadigan:
that's maybe sloppy wording on my part. the vap checks whoever writes the authorino CR (kuadrant-operator's SA), not what authorino itself uses, and only asks for a verb for fields the object actually has set. of the three, kuadrant-operator main only ever sets clusterWide: true (https://github.com/Kuadrant/kuadrant-operator/pull/2220 dropped supersedingHostSubsets, and it's never set image), so set-cluster-wide is the only one that'd fire.

by "released versions" here i meant kuadrant-operator releases, not authorino. that PR isn't in any tag or on release-1.5 yet, so v1.5.3 and earlier still write supersedingHostSubsets: true and would need that verb too if the policy is applied there.

so in short: main only needs set-cluster-wide, but i'd just bind all three since the guide already tells admins to, it covers a 1.5.x backport, and the extras are harmless on main. your call though!

DaliborD45:
@jasonmadigan I thought about it and I agree with you. I think we should grant as least privilege as possible. I will enable set-cluster-wide since it is required and also supersedingHostSubsets, so I do not break older releases.

config/vap/restrict_spec_fields_vap.yaml
guicassolato: The goal is to apply the policy so we enforce a safe default, but in a way the customer can still opt-out.

config/install/manifests.yaml
```diff
@@ -528,6 +528,42 @@ rules:
...
+kind: ClusterRole
+metadata:
+  name: authorino-set-superseding-host-subsets
+rules:
+- apiGroups:
+  - operator.authorino.kuadrant.io
+  resources:
+  - authorinos
+  verbs:
+  - set-superseding-host-subsets
```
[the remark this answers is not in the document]
DaliborD45: No, we do not want to automatically assign ClusterRoles. It should be assigned when VAP will be enabled

config/vap/restrict_spec_fields_vap.yaml
DaliborD45: So should I create RoleBindings rightaway in this PR ?

guicassolato: Not the RoleBindings, those are indeed for the user to decide and apply. We want to apply the VAP and VAPBinding. Thing is I don't know if we can just add them to the manifests. At least OLM may reject them as unsupported kinds. If that's the case, then the Operator itself may need to apply those resources at runtime. Either way, the trickiest part is how to apply VAP and VAPBinding in a way the user is allowed to delete/deactivate them, without the Operator (nor OLM) reconciling them back.

PR #352, opened 2026-08-24
docs: created manual describing fix of privilege escalation vulnerabi…

DaliborD45:
Docs describing a manual for fixing problem, that will be fixed in upstream in https://github.com/Kuadrant/authorino-operator/pull/351

## Summary by CodeRabbit

* **Documentation**
  * Updated the privilege-escalation prevention guide to focus exclusively on Authorino resources.
  * Clarified restrictions on cluster-wide permissions, image values and protected specification fields.
  * Refined verification examples and standardised authorised and unauthorised subject placeholders.
  * Removed outdated AuthConfig policies, role bindings and related verification guidance.
  * Corrected a typo and improved terminology for consistency.


docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -0,0 +1,228 @@
+# User guide: Preventing namespace to cluster privilege escalation via Authorino CR
+
+Some fields on `Authorino` and `AuthConfig` resources reach beyond the namespace they live in. For example, setting `spec.clusterWide: true` on an `Authorino` CR lets it read Secrets in *every* namespace of the cluster. The operator applies these fields without checking whether the person setting them should really have that much access, so anyone allowed to create these resources in a single namespace can quietly gain cluster-wide access.
+
+The two [ValidatingAdmissionPolicies](https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/) below close that gap. They block those fields unless the user has been given a special permission for them, and you hand that permission only to the Kuadrant Operator, which needs it to do its job.
```
guicassolato: Let's keep the phrasing generic so it's not coupled to one specific use of Authorino:

```suggestion
The two [ValidatingAdmissionPolicies](https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/) below close that gap. They block those fields unless the user has been given a special permission for them, and you hand that permission only to the subjects that need it to do their job.
```

docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -0,0 +1,228 @@
...
+The two policies:
+
+| Policy                               | Resource      | Denies                                                                                              | Allowed only for subjects with |
+|--------------------------------------|---------------|----------------------------------------------------------------------------------------------------|--------------------------------|
+| `authorino-restrict-spec-fields`     | `authorinos`  | `spec.image` (any value, blocked outright); `spec.clusterWide: true`                                | `set-cluster-wide` on `authorinos` (for `clusterWide`) |
+| `authconfig-restrict-all-namespaces` | `authconfigs` | `spec.authentication.*.apiKey.allNamespaces: true`; `spec.authentication.*.x509.allNamespaces: true` | `set-privileged-fields` on `authconfigs` |
+
+Follow the three steps below: create the policies, create the ClusterRoles that grant those permissions, then bind the roles to the Kuadrant Operator.
+
+## 1. Create the ValidatingAdmissionPolicies
```
guicassolato: We should probably change the order of the operations to prevent regression:
1. Create the ClusterRoles
2. Create the (Cluster)RoleBindings
3. Create the VAPs

This will prevent authorisation failures for subjects handling the target resources concurrently while the policy is being applied.

docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -0,0 +1,228 @@
...
+
+> [!WARNING]
+> `ValidatingAdmissionPolicy` is GA (enabled by default) in Kubernetes 1.30+ and available as a beta (feature gate `ValidatingAdmissionPolicy` + the `admissionregistration.k8s.io/v1beta1` API) from 1.28. On clusters older than this the policies will not be enforced. Verify your cluster version with `kubectl version`.
+
+The two policies:
+
+| Policy                               | Resource      | Denies                                                                                              | Allowed only for subjects with |
+|--------------------------------------|---------------|----------------------------------------------------------------------------------------------------|--------------------------------|
+| `authorino-restrict-spec-fields`     | `authorinos`  | `spec.image` (any value, blocked outright); `spec.clusterWide: true`                                | `set-cluster-wide` on `authorinos` (for `clusterWide`) |
+| `authconfig-restrict-all-namespaces` | `authconfigs` | `spec.authentication.*.apiKey.allNamespaces: true`; `spec.authentication.*.x509.allNamespaces: true` | `set-privileged-fields` on `authconfigs` |
```
guicassolato: I'm starting to think at this more as a collection of multiple VAPs/ClusterRoles, rather than trying to group them  by resource or similarity of any kind. The goal being to provide fine-grained configuration and establish a more flexible template for [shipping by default with the Operator](https://github.com/Kuadrant/authorino-operator/pull/351) later, when changing the form of the resources becomes harder (may imply migration and/or cleanup effort, etc).

So, even though it may look overkill at first, I wonder if we shouldn't have one separate ClusterRole per feature we want to control access to.

The way it's proposed right now makes it harder to decouple permissions for `authorinos.spec.image` from `authorinos.spec.clusterWide`. The former requires breaking the VAP down while only the latter incorporates RBAC. Similarly, `authconfigs.spec.authentication.apiKey.allNamespaces` and `authconfigs.spec.authentication.x509.allNamespaces` both go with `set-privileged-fields`.

Instead, we could still have 2 VAPs (one for each kind of resource), but one ClusterRole per operation. I.e.:

<table>
<thead>
<tr>
<th>Resource</th>
<th>Field</th>
<th>"Verb"</th>
</tr>
</thead>
<tbody>
<tr>
<td rowspan="2"><code>authorinos</code></td>
<td><code>spec.image</code></td>
<td><code>set-image</code></td>
</tr>
<tr>
<td><code>spec.clusterWide</code></td>
<td><code>cluster-wide</code></td>
</tr>
<tr>
<td rowspan="2"><code>authconfigs</code></td>
<td><code>spec.authentication.apiKey.allNamespaces</code></td>
<td><code>cluster-wide-api-keys</code></td>
</tr>
<tr>
<td><code>spec.authentication.x509.allNamespaces</code></td>
<td><code>cluster-wide-x509-cas</code></td>
</tr>
</tbody>
</table>
 
The validation expressions would all follow the same pattern. E.g.:
- `has(object.spec.clusterWide) && object.spec.clusterWide || authorizer.requestResource.check('cluster-wide').allowed()`
- `has(object.spec.authentication) && object.spec.authentication.exists(k, (has(object.spec.authentication[k].apiKey) && has(object.spec.authentication[k].apiKey.allNamespaces) && object.spec.authentication[k].apiKey.allNamespaces) || authorizer.requestResource.check('cluster-wide-api-keys').allowed()`
- …

docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -0,0 +1,228 @@
+# User guide: Preventing namespace to cluster privilege escalation via Authorino CR
+
+Some fields on `Authorino` and `AuthConfig` resources reach beyond the namespace they live in. For example, setting `spec.clusterWide: true` on an `Authorino` CR lets it read Secrets in *every* namespace of the cluster. The operator applies these fields without checking whether the person setting them should really have that much access, so anyone allowed to create these resources in a single namespace can quietly gain cluster-wide access.
+
+The two [ValidatingAdmissionPolicies](https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/) below close that gap. They block those fields unless the user has been given a special permission for them, and you hand that permission only to the Kuadrant Operator, which needs it to do its job.
+
+> [!WARNING]
+> `ValidatingAdmissionPolicy` is GA (enabled by default) in Kubernetes 1.30+ and available as a beta (feature gate `ValidatingAdmissionPolicy` + the `admissionregistration.k8s.io/v1beta1` API) from 1.28. On clusters older than this the policies will not be enforced. Verify your cluster version with `kubectl version`.
+
```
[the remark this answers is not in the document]
DaliborD45: I am just going to remove this warning since the versions are not even getting security patches, so there is no need to support them.

docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -0,0 +1,228 @@
...
+        apiVersions: ["v1beta1"]
+        operations: ["CREATE", "UPDATE"]
+        resources: ["authorinos"]
+  variables:
+    - name: isExempt
+      expression: "authorizer.requestResource.check('set-cluster-wide').allowed()"
+    - name: wantsClusterWide
+      expression: "has(object.spec.clusterWide) && object.spec.clusterWide"
+    - name: wasClusterWide
+      expression: "request.operation == 'UPDATE' && has(oldObject.spec.clusterWide) && oldObject.spec.clusterWide"
```
guicassolato: Left at https://github.com/Kuadrant/authorino-operator/pull/351/changes#r3843771298 a comment about the need for/risks of checking the old state.

docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -0,0 +1,228 @@
...
+```
+
+## 3. Grant the roles to the Kuadrant Operator
+
+Adjust the ServiceAccount `name` and `namespace` if yours differ from the defaults shown.
+
+```sh
+kubectl apply -f - <<'EOF'
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
```
guicassolato, line 79: I haven't tried the CEL `authorizer` yet but because we're using the `authorizer.requestResource` binding I wonder if it wouldn't work with RoleBindings too (instead of ClusterRoleBinding). My understanding is that the shortcut resolves to the resource [including namespace](https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/#validation-expression). If the authoriser issues a SubjectAccessRequest with all attributes – I would guess it does –, then it should be possible to authorise on a specific namespaces instead of always at cluster scope. If that's the case, perhaps the doc should state that as preferable.

docs/user-guides/preventing-privilege-escalation.md
DaliborD45: I agree with this, also with the point that it makes the permission assignment more fine-grained. However I was thinking about it more like: If I want to block cluster wide to be true, than I also do not want spec.image to be set. In other words, these two policies will always be used together, so there is no point of creating them separately. It could also create a bug where someone would just set one permission restriction, but would forget the second one, so the vulnerability would still be present.

DaliborD45: But if there is a case where we actually need to choose just one (which might be but I am no aware of), than sure I can split the VAPs.

docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -0,0 +1,228 @@
...
+metadata:
+  name: authorino-set-privileged-fields
+rules:
+  - apiGroups: ["authorino.kuadrant.io"]
+    resources: ["authconfigs"]
+    verbs: ["set-privileged-fields"]
+EOF
+```
+
+## 3. Grant the roles to the Kuadrant Operator
```
guicassolato: Like https://github.com/Kuadrant/authorino-operator/pull/352/changes#r3844658226, any specific SA we mention should be provided as an example only. The VAP approach is recommended for all usages. Maybe name this step:


```suggestion
## 3. Grant the roles to a specific ServiceAccount
```

docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -0,0 +1,228 @@
...
+metadata:
+  name: kuadrant-operator-set-cluster-wide
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: authorino-set-cluster-wide
+subjects:
+  - kind: ServiceAccount
+    name: kuadrant-operator-controller-manager
+    namespace: kuadrant-system
```
guicassolato: Using this SA as an example is fine, as long as it's clear it's _only an example_.

As a general principle: dependencies should only flow in one direction. Kuadrant Operator depends on Authorino Operator, so it's natural for it to reference Authorino. The reverse shouldn't happen — Authorino Operator is an upstream project and should have as few references to its downstream consumers as possible, ideally none at all.

docs/user-guides/preventing-privilege-escalation.md
guicassolato: We can keep it at 2 VAPs – one for each resource. Let's split the ClusterRoles into 4.

docs/user-guides/preventing-privilege-escalation.md
DaliborD45: Yes I agree with this. Authorino and authorino-operator are always in the one namespace only, so there is no need to create it cluster-wide role. I will change it to Role and RoleBindinjg

guicassolato: A ClusterRole is fine if you want to keep the changes to a minimum. It's particularly the RoleBinding that I personally think would add another subtle – yet important – dimension to the example: scoping.

docs/user-guides/preventing-privilege-escalation.md
DaliborD45: And do we want a role, where somebody can set spec.image ? Because AFAIK you said that it should be limited to everyone. In that case there would be no reason to create role which would allow that behavior

guicassolato: It's a good question. I think a Role would be helpful.

I know cluster admins could delete/edit the VAP altogether to drop the constraint (for debugging). However, assuming a single VAP per resource kind and the VAP shipped in the future as part of a control plane that could try to keep consistency of it with the original state (e.g.: OLM), a Role for each feature – including `spec.image` – makes it easier to maintain.

It's also simpler in the sense _all_ feature gates follow the same pattern:
- one ClusterRole per feature
- disabled by default (RoleBindings need to be created manually by the user)
- can be enabled via RBAC for individual subjects (by creating the required RoleBindings and ClusterRoleBindings as needed)
- all rules enforced by a single, centralised VAP per resource kind (2 VAPs in total for the entire system).

DaliborD45: Okay, in terms of consistency, it definitely makes sense. I’ll implement it this way.

docs/user-guides/preventing-privilege-escalation.md
DaliborD45: I was thinking about this for a while now, and I have come to the conclusion that we should remove the *was check. If resource already had the permission, than after setting the restriction it should either gain role for access or change the attribute to false/delete it

docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -0,0 +1,262 @@
+# User guide: Preventing namespace to cluster privilege escalation via Authorino CR
```
guicassolato:
```suggestion
# User guide: Preventing namespace to cluster privilege escalation
```

docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -0,0 +1,262 @@
...
+
+Some fields on `Authorino` and `AuthConfig` resources reach beyond the namespace they live in. For example, setting `spec.clusterWide: true` on an `Authorino` CR lets it read Secrets in *every* namespace of the cluster. The operator applies these fields without checking whether the person setting them should really have that much access, so anyone allowed to create these resources in a single namespace can quietly gain cluster-wide access.
+
+The two [ValidatingAdmissionPolicies](https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/) below close that gap. They block those fields unless the user has been given a special permission for them, and you hand that permission only to the subjects that need it to do their job.
+
+The two policies:
+
+| Policy                               | Resource      | Denies                                                                                              | Allowed only for subjects with |
```
guicassolato:
```suggestion
| Policy                               | Resource      | Denies                                                                                              | Required Roles to allow |
```

docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -0,0 +1,262 @@
...
+|--------------------------------------|---------------|----------------------------------------------------------------------------------------------------|--------------------------------|
+| `authorino-restrict-spec-fields`     | `authorinos`  | `spec.image` (any value); `spec.clusterWide: true`                                                  | `set-image` on `authorinos` (for `image`); `set-cluster-wide` on `authorinos` (for `clusterWide`) |
+| `authconfig-restrict-all-namespaces` | `authconfigs` | `spec.authentication.*.apiKey.allNamespaces: true`; `spec.authentication.*.x509.allNamespaces: true` | `set-apikey-all-namespaces` (for `apiKey.allNamespaces`); `set-x509-all-namespaces` (for `x509.allNamespaces`) on `authconfigs` |
```
guicassolato:
```suggestion
<table>
  <thead>
    <tr>
      <th>Policy</th>
      <th>Resource</th>
      <th>Denies</th>
      <th>Role required to allow</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td rowspan="2"><code>authorino-restrict-spec-fields</code></td>
      <td rowspan="2"><code>authorinos</code></td>
      <td><code>spec.image</code> (any value)</td>
      <td><code>set-image</code> on <code>authorinos</code></td>
    </tr>
    <tr>
      <td><code>spec.clusterWide: true</code></td>
      <td><code>set-cluster-wide</code> on <code>authorinos</code></td>
    </tr>
    <tr>
      <td rowspan="2"><code>authconfig-restrict-all-namespaces</code></td>
      <td rowspan="2"><code>authconfigs</code></td>
      <td><code>spec.authentication.*.apiKey.allNamespaces: true</code></td>
      <td><code>set-apikey-all-namespaces</code> on <code>authconfigs</code></td>
    </tr>
    <tr>
      <td><code>spec.authentication.*.x509.allNamespaces: true</code></td>
      <td><code>set-x509-all-namespaces</code> on <code>authconfigs</code></td>
    </tr>
  </tbody>
</table>
```

docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -0,0 +1,262 @@
...
+rules:
+  - apiGroups: ["authorino.kuadrant.io"]
+    resources: ["authconfigs"]
+    verbs: ["set-x509-all-namespaces"]
+EOF
+```
+
+## 2. Grant the roles to a specific ServiceAccount
+
+Replace the `<...>` placeholders (`<sa-name>`, `<namespace-of-sa>`, `<authorino-namespace>`, `<authconfig-namespace>`) with your ServiceAccount and namespaces.
```
guicassolato:
```suggestion
Use the RoleBindings below as a template to grant access to your ServiceAccounts and Users. Pick on the RoleBindings that make sense for your use case. Replace the placeholders (`<sa-name>`, `<namespace-of-sa>`, `<authorino-namespace>`, `<authconfig-namespace>`) with the appropriate values.
```

docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -0,0 +1,262 @@
...
+        resources: ["authorinos"]
+  variables:
+    - name: isExemptClusterWide
+      expression: "authorizer.requestResource.check('set-cluster-wide').allowed()"
+    - name: isExemptImage
+      expression: "authorizer.requestResource.check('set-image').allowed()"
+    - name: wantsClusterWide
+      expression: "has(object.spec.clusterWide) && object.spec.clusterWide"
+    - name: wantsImage
+      expression: "has(object.spec.image) && object.spec.image != ''"
```
guicassolato, line 193: We're using each variable only once in the validations. Do we still need to define them?

docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -0,0 +1,262 @@
...
+  name: authorino-set-x509-all-namespaces
+  namespace: <authconfig-namespace>
+rules:
+  - apiGroups: ["authorino.kuadrant.io"]
+    resources: ["authconfigs"]
+    verbs: ["set-x509-all-namespaces"]
+EOF
+```
+
+## 2. Grant the roles to a specific ServiceAccount
```
guicassolato: Sorry, I know I suggested the current title for the section. To make it even more generic, maybe:

```suggestion
## 2. Grant the access to the restricted fields
```

docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -0,0 +1,262 @@
...
+
+The two policies:
+
+| Policy                               | Resource      | Denies                                                                                              | Allowed only for subjects with |
+|--------------------------------------|---------------|----------------------------------------------------------------------------------------------------|--------------------------------|
+| `authorino-restrict-spec-fields`     | `authorinos`  | `spec.image` (any value); `spec.clusterWide: true`                                                  | `set-image` on `authorinos` (for `image`); `set-cluster-wide` on `authorinos` (for `clusterWide`) |
+| `authconfig-restrict-all-namespaces` | `authconfigs` | `spec.authentication.*.apiKey.allNamespaces: true`; `spec.authentication.*.x509.allNamespaces: true` | `set-apikey-all-namespaces` (for `apiKey.allNamespaces`); `set-x509-all-namespaces` (for `x509.allNamespaces`) on `authconfigs` |
+
+Follow the three steps below: create the Roles that grant those permissions, bind the roles to the specific SA, then create the policies.
```
guicassolato:
```suggestion
Follow the three steps below: create the Roles that grant those permissions, bind the roles to specific SAs and Users, then apply the policies.
```

guicassolato:
Only a few nits now, but looking great!

docs/user-guides/preventing-privilege-escalation.md
DaliborD45: No we do not, but it seems much more clear to me, since the expression is quite long and chaining them would create quite a long condition

docs/user-guides/preventing-privilege-escalation.md
DaliborD45: Actually didnt know that markdown allows to use html format, nice !

docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -0,0 +1,289 @@
...
+</table>
+
+Follow the three steps below: create the Roles that grant those permissions, bind the roles to specific SAs and Users, then apply the policies.
+
+## 1. Create the Roles
+
+```sh
+kubectl apply -f - <<'EOF'
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
```
guicassolato: Can we please change these to ClusterRoles so they don't need to be repeated at every Authorino/AuthConfig namespace where we want control access to the features?

docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -0,0 +1,289 @@
...
+  namespace: <authconfig-namespace>
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: authorino-set-x509-all-namespaces
+subjects:
+  - kind: ServiceAccount
+    name: <sa-name>
+    namespace: <namespace-of-sa>
+EOF
```
guicassolato, line 164: While testing the procedure, I noticed the Operator needs to be granted the `set-image` and `set-cluster-wide` permissions or it will fail to manage Authorino CRs for users with those permissions. IOW, we need to add to the instructions these additional 2 ClusterRoleBindings:

```sh
kubectl apply -f -<<EOF
kind: ClusterRoleBinding
metadata:
  name: authorino-set-cluster-wide
roleRef:
  kind: ClusterRole
  name: authorino-set-cluster-wide
subjects:
- kind: ServiceAccount
  name: authorino-operator
  namespace: authorino-operator # replace with the namespace where the operator is running
---
kind: ClusterRoleBinding
metadata:
  name: authorino-set-image
roleRef:
  kind: ClusterRole
  name: authorino-set-image
subjects:
- kind: ServiceAccount
  name: authorino-operator
  namespace: authorino-operator # replace with the namespace where the operator is running
EOF
```

guicassolato: Same for the Authorino instances. They must be granted `authorino-set-apikey-all-namespaces` and `authorino-set-x509-all-namespaces` permissions at the same scope the instance operates (cluster-wide or namespace).

DaliborD45: So, do you mean that the authorino-operator also needs these four roles assigned to it? I’m not quite sure I understand why. Is it because it needs to perform some kind of self-checking?

guicassolato: That is what I observed when testing the procedure. Step "4. Verifying the VAPs" succeeds, but it hides the fact that the requested Authorino instances are never deployed. Instead, the Operator fails due to missing these permissions.

I then _assumed_ the Authorino instance would stumble into similar problem.

I'll try the procedure again from scratch and paste the exact error message here I see in the logs, unless you tell me you did that already and the Authorino instances were successfully deployed after step 4 without any authorisation hiccup perceived in the Operator's logs.

guicassolato: Confirmed. Here's the error in the Operator logs when the Operator's SA misses the permissions to handle `Authorino` CRs with `clusterWide: true`:

```
{"level":"error","ts":"2026-08-26T09:12:59Z","logger":"authorino-operator.controller.Authorino","msg":"Reconciler error","controller":"authorino","controllerGroup":"operator.authorino.kuadrant.io","controllerKind":"Authorino","Authorino":{"name":"authorino","namespace":"infra"},"namespace":"infra","name":"authorino","reconcileID":"3315fa5a-12e0-48f0-9ca9-d3f2ff0ff5c9","error":"authorinos.operator.authorino.kuadrant.io \"authorino\" is forbidden: ValidatingAdmissionPolicy 'authorino-restrict-spec-fields' with binding 'authorino-restrict-spec-fields-binding' denied request: spec.clusterWide: true can only be set by a subject granted the 'set-cluster-wide' permission on authorinos","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler\n\t/opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.23.3/pkg/internal/controller/controller.go:495\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem\n\t/opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.23.3/pkg/internal/controller/controller.go:438\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func1.1\n\t/opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.23.3/pkg/internal/controller/controller.go:313"}
```

guicassolato: Similarly, for the Authorino instance when it tries to reconcile an AuthConfig with `apiKey.allNamespaces: true`:

```
{"level":"error","ts":"2026-08-26T09:16:53Z","logger":"authorino.controller-runtime.manager.controller.authconfig.statusupdater","msg":"failed to update the resource","authconfig":{"name":"infra-apikey-auth","namespace":"infra"},"error":"authconfigs.authorino.kuadrant.io \"infra-apikey-auth\" is forbidden: User \"system:serviceaccount:infra:authorino-authorino\" cannot update resource \"authconfigs/status\" in API group \"authorino.kuadrant.io\" in the namespace \"infra\"","stacktrace":"github.com/kuadrant/authorino/controllers.(*AuthConfigStatusUpdater).updateAuthConfigStatus\n\t/usr/src/authorino/controllers/auth_config_status_updater.go:162\ngithub.com/kuadrant/authorino/controllers.(*AuthConfigStatusUpdater).Reconcile\n\t/usr/src/authorino/controllers/auth_config_status_updater.go:81\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile\n\t/opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:119\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:316\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:266\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:227"}
```

docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -0,0 +1,524 @@
...
+### A normal user is blocked
+
+Try to create resources that break the rules. Run these as a regular user (one *without* the permissions) — both should be **rejected**:
+
+> [!NOTE]
+> Don't run these as a cluster admin. Anything with wildcard access (`verbs: ["*"]`) — which cluster admins have — satisfies the `set-cluster-wide` / `set-image` / `set-apikey-all-namespaces` / `set-x509-all-namespaces` checks and is treated as exempt, so the request would be **allowed** and a real cluster-wide instance created. Use an ordinary user (or `--as=some-user`) to see the policies block.
+
+```sh
+# Authorino with clusterWide: true — should be DENIED
+kubectl apply -f - <<'EOF'
```
guicassolato:
```suggestion
kubectl apply --as=<unauthorized-subject> -f - <<'EOF'
```

docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -0,0 +1,524 @@
...
+      enabled: false
+  oidcServer:
+    tls:
+      enabled: false
+EOF
+```
+
+```sh
+# AuthConfig with allNamespaces: true — should be DENIED
+kubectl apply -f - <<'EOF'
```
guicassolato:
```suggestion
kubectl apply --as=<unauthorized-subject> -f - <<'EOF'
```

docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -0,0 +1,524 @@
...
+... is forbidden: ValidatingAdmissionPolicy 'authorino-restrict-spec-fields' ... denied request: spec.clusterWide: true can only be set by a subject granted the 'set-cluster-wide' permission on authorinos
+```
+
+### A permitted subject is allowed
+
+Now run the same requests as a subject that holds the matching permission (granted in steps 1–2). Both should be **admitted**. Replace `<permitted-subject>` with the subject you granted the permission to (e.g. `system:serviceaccount:<namespace>:<sa>`):
+
+```sh
+# Authorino with clusterWide: true, as a subject granted 'set-cluster-wide' — should be ALLOWED
+kubectl apply --as=<permitted-subject> -f - <<'EOF'
```
guicassolato: nit: for consistency with my previous suggestion only.

```suggestion
kubectl apply --as=<authorized-subject> -f - <<'EOF'
```

docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -0,0 +1,524 @@
...
+      enabled: false
+  oidcServer:
+    tls:
+      enabled: false
+EOF
+```
+
+```sh
+# AuthConfig with allNamespaces: true, as a subject granted 'set-apikey-all-namespaces' — should be ALLOWED
+kubectl apply --as=<permitted-subject> -f - <<'EOF'
```
guicassolato:
```suggestion
kubectl apply --as=<authorized-subject> -f - <<'EOF'
```

docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -0,0 +1,524 @@
...
+EOF
+```
+
+### Updates are re-checked, not just creates
+
+Because the policies match `UPDATE` as well as `CREATE`, they re-evaluate on every change. A subject without the permission can still edit **unrelated** fields of a compliant resource, but is blocked the moment it tries to switch a restricted field on. Using the namespaced instance created above:
+
+```sh
+# Change an unrelated field (logLevel) on the namespaced Authorino — should be ALLOWED
+kubectl apply --as=<subject> -f - <<'EOF'
```
guicassolato:
```suggestion
kubectl apply --as=<unauthorized-subject> -f - <<'EOF'
```

docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -0,0 +1,524 @@
...
+      enabled: false
+  oidcServer:
+    tls:
+      enabled: false
+EOF
+```
+
+```sh
+# Flip the same instance to clusterWide: true — should be DENIED
+kubectl apply --as=<subject> -f - <<'EOF'
```
guicassolato:
```suggestion
kubectl apply --as=<unauthorized-subject> -f - <<'EOF'
```

docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -0,0 +1,524 @@
...
+EOF
+```
+
+### Permissions bound with a RoleBinding are namespace-scoped
+
+The exemption check runs against the namespace of the resource being admitted. If you grant the permission with a `RoleBinding` (rather than a `ClusterRoleBinding`), the subject is exempt only in that namespace.
+
+```sh
+# Subject granted 'set-cluster-wide' via a RoleBinding in <namespace-a> — should be ALLOWED
+kubectl apply --as=<permitted-subject> -f - <<'EOF'
```
guicassolato:
```suggestion
kubectl apply --as=<authorized-subject> -f - <<'EOF'
```

docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -0,0 +1,524 @@
...
+      enabled: false
+  oidcServer:
+    tls:
+      enabled: false
+EOF
+```
+
+```sh
+# Same subject, same request, in <namespace-b> where it has no binding — should be DENIED
+kubectl apply --as=<permitted-subject> -f - <<'EOF'
```
guicassolato:
```suggestion
kubectl apply --as=<authorized-subject> -f - <<'EOF'
```

docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -0,0 +1,524 @@
...
+
+Now run the same requests as a subject that holds the matching permission (granted in steps 1–2). Both should be **admitted**. Replace `<permitted-subject>` with the subject you granted the permission to (e.g. `system:serviceaccount:<namespace>:<sa>`):
+
+```sh
+# Authorino with clusterWide: true, as a subject granted 'set-cluster-wide' — should be ALLOWED
+kubectl apply --as=<permitted-subject> -f - <<'EOF'
+apiVersion: operator.authorino.kuadrant.io/v1beta1
+kind: Authorino
+metadata:
+  name: policy-test
```
guicassolato: We should use different names for each `Authorino` and `AuthConfig` in the examples, to avoid people running the procedure end up testing the permissions enforced on an UPDATE instead of a CREATE.

docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -0,0 +1,524 @@
...
+EOF
+```
+
+```sh
+# AuthConfig with allNamespaces: true, as a subject granted 'set-apikey-all-namespaces' — should be ALLOWED
+kubectl apply --as=<permitted-subject> -f - <<'EOF'
+apiVersion: authorino.kuadrant.io/v1beta3
+kind: AuthConfig
+metadata:
+  name: policy-test
```
guicassolato: Can we please use more suggestive names for the `Authorino` CRs and `AuthConfig` CRs in the example, instead of `policy-test` for both? E.g.: `authorino-cluster-wide-1`, `authorino-cluster-wide-2`, `authorino-namespaced-1`, `policy-all-namespaces-1`, `policy-namespaced-1`, etc

docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -0,0 +1,524 @@
...
+      enabled: false
+  oidcServer:
+    tls:
+      enabled: false
+EOF
+```
+
+```sh
+# AuthConfig with allNamespaces: false — should be ALLOWED for any subject
+kubectl apply --as=<subject> -f - <<'EOF'
```
guicassolato:
```suggestion
kubectl apply --as=<unauthorized-subject> -f - <<'EOF'
```

docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -0,0 +1,524 @@
...
+EOF
+```
+
+### Resources without the restricted fields are always allowed
+
+The policies only look at the restricted fields. A resource that leaves them unset (or `false`) is admitted for **any** subject, whether or not it holds a permission:
+
+```sh
+# Namespaced Authorino (clusterWide omitted) — should be ALLOWED for any subject
+kubectl apply --as=<subject> -f - <<'EOF'
```
guicassolato:
```suggestion
kubectl apply --as=<unauthorized-subject> -f - <<'EOF'
```

docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -0,0 +1,524 @@
...
+# AuthConfig with allNamespaces: false — should be ALLOWED for any subject
+kubectl apply --as=<subject> -f - <<'EOF'
+apiVersion: authorino.kuadrant.io/v1beta3
+kind: AuthConfig
+metadata:
+  name: policy-test-namespaced
+  namespace: <namespace>
+spec:
+  hosts:
+    - test.example.com
```
guicassolato: Using the same hostnames at all AuthConfigs, with the AuthConfigs reconciled by the same Authorino instance, will prevent the resource to move to a `Ready` state. Although this is not exactly an issue for testing the VAPs, readers may get confused if they check the status after the operation.

If the Authorino instance's SA lacks permission to handle the AuthConfig (i.e., user missed or failed at step 2), then the AuthConfig also won't move to a `Ready` (though for a different reason). This would add to the confusion, possibly mistaking one case for the other.

I'd use unique hostnames for the AuthConfigs just in case.

DaliborD45:
@guicassolato I squashed all commits into one, since it was just fixing comments and some of them were not signed off

guicassolato (dismissed):

guicassolato:
I think we should add `supersedingHostSubsets: true` to the scope here. It's another "dangerous value" that we should recommend reserving to privileged users only.

Please feel free to include in this PR for leave it for a separate one.

DaliborD45:
@guicassolato added that supersedingHostSubsets role

bundle/manifests/authorino-operator.clusterserviceversion.yaml
```diff
@@ -55,7 +55,7 @@ metadata:
     capabilities: Basic Install
     categories: Integration & Delivery
     containerImage: quay.io/kuadrant/authorino-operator:latest
-    createdAt: "2026-07-10T12:38:29Z"
+    createdAt: "2026-09-10T09:01:43Z"
```
guicassolato, line 58: This doesn't belong to this PR. You probably need a rebase instead now that #358 has been merged.

DaliborD45: Well I had to regenerated manifests, otherwise the pipeline was failling

DaliborD45: @guicassolato I have rebased the branch from main and it removed the commit with generating manifests automatically, so now everyone should be okay.

guicassolato (dismissed):
I think this PR is ready to merge IMO. I'm leaving a few enhancement suggestions, mostly inspired by late additions to its sibling PR in the Authorino repo https://github.com/Kuadrant/authorino/pull/678, but none are blockers.

1. Maybe add a pre-rollout audit snippet so admins know which existing CRs will become read-only before they apply the policy. E.g.:

```
kubectl get authorinos -A -o json | jq -r '.items[]
  | select((.spec.clusterWide // false) or (.spec.supersedingHostSubsets // false) or ((.spec.image // "") != ""))
  | "\(.metadata.namespace)/\(.metadata.name)"'
```

2. Generalize the operator-SA warning to "controllers and GitOps agents" — a CI SA reconciling an Authorino CR that pins `spec.image` will be blocked unless granted `set-image`, just like an operator would. This is a realistic break the current doc doesn't call out.

3. State the Kubernetes prerequisite: 1.30+ to use `admissionregistration.k8s.io/v1`.

DaliborD45:
@guicassolato I have resolved your suggestions and applied them to the PR, everything should be fine now, I just need new approve so I can merge it.

guicassolato (approved):
LGTM. Thanks for the hard work @DaliborD45!

guicassolato:
cc @jasonmadigan

PR #363, opened 2026-09-17
fix: make authorization service headless to enable gRPC load distribution

### LINKED ISSUE #362: Make authorization service headless to enable gRPC load distribution across replicas
## Problem

When Authorino is deployed with multiple replicas, the `<name>-authorino-authorization` Kubernetes Service is a standard ClusterIP service. The Kuadrant operator's EnvoyFilter configures Envoy with a `STRICT_DNS` cluster pointing to this service's FQDN. Because a ClusterIP service resolves to a single virtual IP, Envoy creates one upstream endpoint and maintains one persistent HTTP/2 connection to it. All multiplexed gRPC ext_authz streams ride this single connection, which kube-proxy forwards to the same Authorino pod regardless of replica count.

**Observed behavior with 3 Authorino replicas under load:**
- One pod received zero requests across a full load test
- The other two received ~250 and ~238 requests respectively
- Above ~25 concurrent requests, `UNAVAILABLE` (gRPC code 14) errors appear despite healthy replicas being available

## Root Cause

```
STRICT_DNS cluster → ClusterIP FQDN → single VIP → single upstream host
  → single persistent HTTP/2 connection → all gRPC streams pinned to 1–2 pods
  → ROUND_ROBIN lb_policy has nothing to round-robin against
```

## Fix

Make the `authorino-authorization` service headless (`clusterIP: None`). DNS then returns one A record per pod IP, giving Envoy N upstream hosts for N replicas. `ROUND_ROBIN` distributes each new gRPC stream across N connections (one per pod), achieving true stream-level load distribution.

The reconciler must handle the ClusterIP immutability constraint: an existing ClusterIP service cannot be patched to headless — it must be deleted and recreated.

## Companion fix

kuadrant/kuadrant-operator — adds `dns_refresh_rate` and `outlier_detection` to the ext_authz cluster so pod IP churn during rolling updates is handled gracefully.

thomasmaas:
## Problem

When Authorino is deployed with multiple replicas, the \`<name>-authorino-authorization\` Kubernetes Service is a standard ClusterIP service. The Kuadrant operator's EnvoyFilter configures Envoy with a \`STRICT_DNS\` cluster pointing to this service's FQDN.

Because a ClusterIP service resolves to a single virtual IP, Envoy creates **one upstream endpoint** and maintains **one persistent HTTP/2 connection** to it. All multiplexed gRPC ext_authz streams ride this single connection, which kube-proxy forwards to the same Authorino pod regardless of replica count. \`ROUND_ROBIN\` lb_policy has nothing to round-robin against.

**Observed with 3 Authorino replicas under load:**
- One pod received zero requests across a full 1000-request load test
- The remaining two split all traffic (~250 / ~238)
- Above ~25 concurrent requests, \`UNAVAILABLE\` (gRPC code 14) errors appear despite healthy replicas being idle

## Precedent

The Limitador service (\`limitador-operator/pkg/limitador/k8s_objects.go\`) has been headless (\`ClusterIP: None\`) since its introduction, for exactly this reason — DNS returns pod IPs directly, giving Envoy one upstream host per replica and enabling ROUND_ROBIN to distribute gRPC streams across all pods. This PR brings the Authorino authorization service into alignment with that established pattern.

## Fix

Make the \`authorino-authorization\` service headless (\`clusterIP: None\`). DNS then returns one A record per pod IP. Envoy's \`STRICT_DNS\` cluster creates one upstream host per replica. \`ROUND_ROBIN\` distributes each new gRPC stream across N connections (one per pod), achieving ~1/N traffic per replica.

\`\`\`
Before: STRICT_DNS → ClusterIP VIP → 1 upstream host → 1 HTTP/2 connection → 1 pod
After:  STRICT_DNS → pod IPs (N) → N upstream hosts → N HTTP/2 connections → all pods
\`\`\`

## Changes

- **\`pkg/resources/k8s_services.go\`**: Set \`ClusterIP: "None"\` on the auth service. OIDC and metrics services are unchanged.
- **\`pkg/reconcilers/authorino_reconciler.go\`**: Add \`isClusterIPImmutableError\` helper and handling in \`reconcileService\` to detect the 422 Invalid error returned by SSA when migrating an existing ClusterIP service to headless, and perform a delete+recreate.

## Upgrade path

On upgrade, the reconciler detects the immutable ClusterIP conflict and deletes+recreates the service. During the recreation window (typically < 5 seconds), ext_authz calls may fail. Plan upgrades during low-traffic periods.

## Testing

- \`pkg/reconcilers\` and \`pkg/resources\` unit tests pass unchanged
- Integration test (\`controllers/authorino_controller_test.go\`) verifies service creation; the headless field is set correctly on the new service

## Companion

[kuadrant/kuadrant-operator#2281](https://github.com/Kuadrant/kuadrant-operator/pull/2281) — adds \`dns_refresh_rate\` and \`outlier_detection\` to the ext_authz cluster so pod IP churn during rolling updates is handled gracefully.

Closes #362

## Summary by CodeRabbit

- **New Features**
  - Authorino’s authorisation Service now uses headless networking, allowing DNS to resolve individual pod addresses instead of a single virtual IP.
  - Existing Services can be migrated automatically to the new headless configuration.

- **Bug Fixes**
  - Improved recovery when Kubernetes prevents an in-place ClusterIP change, ensuring the Service is recreated with the correct configuration rather than leaving the update failed.


guicassolato (approved):
I haven’t tried it but code LGTM.

guicassolato:
In time – the best-effort recreation of the authorization service as a headless service works for upgrades but there's no downgrade path. One who tries to downgrade Authorino Operator will likely end up with a broken deployment.

guicassolato (changes requested):
Hmm. Maybe I approved the PR a little too soon.

I've tried the migration path locally by first creating an Authorino CR on the `main` branch, then running the Operator with the changes. Here's what I see in the logs and the service is never recreated:

```
2026-09-21T16:58:32+02:00	ERROR	authorino-operator.controller.Authorino	Reconciler error	{"controller": "authorino", "controllerGroup": "operator.authorino.kuadrant.io", "controllerKind": "Authorino", "Authorino": {"name":"authorino","namespace":"default"}, "namespace": "default", "name": "authorino", "reconcileID": "d609d5b2-c1e5-4527-9d70-e0dfe3fc877a", "error": "failed to update authorino-authorino-authorization service, err: Service \"authorino-authorino-authorization\" is invalid: spec.clusterIPs[0]: Invalid value: [\"None\"]: may not change once set"}
```

pkg/reconcilers/authorino_reconciler.go
```diff
@@ -482,6 +504,24 @@ func (r *AuthorinoReconciler) reconcileService(ctx context.Context, desired *k8s
...
+func isClusterIPImmutableError(err error) bool {
+	statusErr, ok := err.(*k8serrors.StatusError)
+	if !ok {
+		return false
+	}
+	if statusErr.ErrStatus.Details == nil {
+		return false
+	}
+	for _, cause := range statusErr.ErrStatus.Details.Causes {
+		if cause.Field == "spec.clusterIP" {
```
guicassolato: I think I may have found the root cause of the [reconciliation error](https://github.com/Kuadrant/authorino-operator/pull/363#pullrequestreview-5268195801):

```suggestion
		if strings.HasPrefix(cause.Field, "spec.clusterIP") {
```

thomasmaas: Good catch — the API server uses `spec.clusterIPs[0]` (plural, indexed), not `spec.clusterIP`. Applied your suggestion (`strings.HasPrefix`) and updated the test to inject the actual field name so this is exercised.

thomasmaas:
On the downgrade concern: if someone downgrades to a pre-headless operator version, the auth service stays headless (the old operator's SSA desired spec omits `ClusterIP` entirely — SSA doesn't remove fields it doesn't manage, so headless is preserved). The old operator should continue to reconcile the service correctly since it only manages ports/selector. That said, if the old operator does assert on the service having a concrete ClusterIP somewhere, that would break. Worth adding a note to the release notes that this change is one-way for the auth service.

pkg/reconcilers/authorino_reconciler.go
thomasmaas: Thanks for catching that Gui

guicassolato (approved):
LGTM.

